PR #17, opened 2018-11-24
Stack_dump on KernelStack and UserLand stack

Orycterope:
We now treat KernelStack dumping and UserLand stack dumping
as two completely different situations, and make appropriate checks
in each one.


kernel/src/i386/stack.rs
```diff
@@ -138,68 +138,108 @@ impl KernelStack {
...
+/// a KernelStack, check that every page of what we're expecting to be a stack is mapped,
+/// and dump it.
+///
+/// Otherwise, it will get the UserLand mapping that `esp` falls in, and dump only in this mapping.
+///
+/// # Safety
+///
+/// This function is "relatively" safe. It checks whether the stack is properly mapped
+/// before attempting to access it. It does create a &[u8] from a technically "shared"
+/// resource. However, the compiler shouldn't know about this, so UB shouldn't be met.
```
roblabla: While I understand the sentiment, this is still technically UB. I think the relevant reason why UB will not happen is that, when using this function, we’re usually dumping our stack. Which should be immutable for the duration of the dump_stack function.

I’m not sure how safe it would be to do this to dump another process’ (live) stack. Probably very unsafe.

Orycterope: I agree, and I found out the hard way that dumping a UserLand stack is not only a bad idea because of safety as you pointed out, it also is useless to our kernel right now. Since in an exception handler we don't have access to the saved $ebp (abi interrupt-x86), we don't know where the bottom of the UserLand stack frame is, so there is no way for us to dump it.

I'm thinking more and more about forbidding UserLand stack dumping altogether, and allow only KernelStack dumping of our current stack, as you mentioned.

Would you be ok with that ?

PR #25, opened 2018-11-27
Warn on unused

Orycterope:
First commit makes the kernel no longer `#![allow(unused)]`.
We still allow `dead_code`, `unused_unsafe` and `unreachable_code`.

All other commits fix all the warnings it generated.


roblabla:
Waaat

roblabla:
Oh it's ok. I think the merge commit did this.

PR #28, opened 2018-11-28
Implement QueryMemory

roblabla:
First implementation of QueryMemory. It is lacking two things:

- PageInfo, which I need to RE from Horizon/NX
- The paging's MappingType needs to be enriched to properly map to MemoryType

### DIFF
--- a/kernel/src/interrupts/syscalls.rs
+++ b/kernel/src/interrupts/syscalls.rs
@@ -3,7 +3,7 @@
 use i386;
 use mem::{VirtualAddress, PhysicalAddress};
 use mem::{UserSpacePtr, UserSpacePtrMut};
-use paging::{MappingFlags, process_memory::QueryMemory};
+use paging::{MappingFlags};
 use frame_allocator::PhysicalMemRegion;
 use process::{Handle, ThreadStruct, ProcessStruct};
 use event::{self, Waitable};
@@ -16,7 +16,7 @@ use alloc::vec::Vec;
 use ipc;
 use super::check_thread_killed;
 use error::UserspaceError;
-use kfs_libkern::{nr, SYSCALL_NAMES, MemoryInfo, MemoryType, MemoryAttributes, MemoryPermissions};
+use kfs_libkern::{nr, SYSCALL_NAMES, MemoryInfo, MemoryAttributes};
 
 extern fn ignore_syscall(nr: usize) -> Result<(), UserspaceError> {
     // TODO: Trigger "unknown syscall" signal, for userspace signal handling.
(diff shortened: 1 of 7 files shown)


kernel/src/paging/mapping.rs
```diff
@@ -41,6 +42,23 @@ pub enum MappingType {
...
+impl<'a> From<&'a MappingType> for kfs_libkern::MemoryType {
+    fn from(ty: &'a MappingType) -> kfs_libkern::MemoryType {
+        match ty {
+            // TODO: Extend MappingType to cover all MemoryTypes
+            // BODY: Currently, MappingType only covers a very limited view of the mappings.
+            // It should have the ability to understand all the various kind of memory allocations,
+            // such as "Heap", "CodeMemory", "SharedMemory", "TransferMemory", etc...
+
+            MappingType::Available => kfs_libkern::MemoryType::Unmapped,
+            MappingType::Guarded => kfs_libkern::MemoryType::Unmapped,
```
Orycterope: Guarded should be "Reserved", not unmapped, as it would cause userspace to ask to allocate in it.

kernel/src/interrupts/syscalls.rs
```diff
@@ -231,6 +231,35 @@ fn create_port(max_sessions: u32, _is_light: bool, _name_ptr: UserSpacePtr<[u8;
     Ok((clienthnd as _, serverhnd as _))
 }
 
+#[inline(never)]
+fn query_memory(mut meminfo: UserSpacePtrMut<MemoryInfo>, _unk: usize, addr: usize) -> Result<usize, UserspaceError> {
+    let curproc = scheduler::get_current_process();
+    *meminfo = match curproc.pmemory.lock().query_memory(VirtualAddress(addr))? {
+        QueryMemory::Available(mapping) => MemoryInfo {
```
Orycterope: do a QueryMemory::get_ref(&self) -> &Mapping, to avoid having to match twice.

(Also rename QueryMemory struct as it's a horrible name)

libuser/src/lib.rs
```diff
@@ -41,6 +41,25 @@ fn init_heap() {
     }
 }
 
+pub fn find_free_address(size: usize, align: usize) -> Result<usize, usize> {
+    let mut addr = 0;
+    // Go over the address space.
+    loop {
+        let (meminfo, _) = syscalls::query_memory(addr)?;
+        if meminfo.memtype == kfs_libkern::MemoryType::Unmapped {
+            let alignedbaseaddr = kfs_libutils::align_up(meminfo.baseaddr, align);
```
Orycterope: This could cause overflow if `align` is unrealistically big.

libuser/src/lib.rs
```diff
@@ -47,17 +47,14 @@ pub fn find_free_address(size: usize, align: usize) -> Result<usize, usize> {
     loop {
         let (meminfo, _) = syscalls::query_memory(addr)?;
         if meminfo.memtype == kfs_libkern::MemoryType::Unmapped {
-            let alignedbaseaddr = kfs_libutils::align_up(meminfo.baseaddr, align);
+            let alignedbaseaddr = kfs_libutils::align_up_checked(meminfo.baseaddr, align).ok_or(1usize)?;
+
             let alignment = alignedbaseaddr - meminfo.baseaddr;
-            if alignment + size < meminfo.size {
+            if alignment.checked_add(size).ok_or(1usize)? < meminfo.size {
```
Orycterope: It should be alignement + *(size - 1)* < meminfo.size

A size of 0 is absurd, this function should check this case and return an error.

Currently, if we want to allocate 0x1000, aligned on page size, and we're considering a hole of exactly 1 page at 0xc000, 
we have
```
alignedbaseaddr = 0xc000
alignement (left padding) = 0
0 + 0x1000 < 0x1000 : false
```
or there was enough space .

Note: We can neither do `addr + size <= mapping.size`, as it would overflow for the very last page, but there was enough space.

PR #30, opened 2018-11-29
Implement SharedMemory

roblabla:
This PR depends on #28 . **DO NOT MERGE BEFORE MERGING #28**

- Implements Create/Map/Unmap shared memory syscalls in kernelspace
- Implements the equivalent syscall in userspace
- Add SharedMemory and MappedSharedMemory userspace types.
- MappedSharedMemory contains some unsafe functions that should probably not
  be used...

### DIFF
--- a/kernel/src/interrupts/syscalls.rs
+++ b/kernel/src/interrupts/syscalls.rs
@@ -232,6 +232,14 @@ fn create_port(max_sessions: u32, _is_light: bool, _name_ptr: UserSpacePtr<[u8; 
 }
 
 fn create_shared_memory(size: u32, _myperm: u32, _otherperm: u32) -> Result<usize, UserspaceError> {
+    // TODO: Frame Allocator should take lengths in bytes.
+    // BODY: The frame allocator currently takes frame numbers instead of byte
+    // lengths. This forces all the error handling upon the callers, which is a
+    // bit of a pain. Instead, it should take byte lengths, and handle the error
+    // cases itself.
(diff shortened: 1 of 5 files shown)


kernel/src/interrupts/syscalls.rs
```diff
@@ -231,6 +231,58 @@ fn create_port(max_sessions: u32, _is_light: bool, _name_ptr: UserSpacePtr<[u8;
     Ok((clienthnd as _, serverhnd as _))
 }
 
+fn create_shared_memory(size: u32, _myperm: u32, _otherperm: u32) -> Result<usize, UserspaceError> {
+    let frames = FrameAllocator::allocate_frames_fragmented((size as usize) / PAGE_SIZE)?;
```
Orycterope: shouldn't this be size aligned *up* to PAGE_SIZE ?

Orycterope:
So, if I understand correctly, closing a SharedMemory handle leaves it mapped, but makes it impossible for the process to unmap it. The SharedMemory is not dropped, because an Arc remains in the process' bookkeeping.
Is that it ?

roblabla:
@Orycterope no. When you map a SharedMemory, MappedSharedMemory gets ownership of the underlying handle. As such, the handle cannot be closed until the MappedSharedMemory gets unmapped (which will close the handle).

kernel/src/interrupts/syscalls.rs
roblabla: Welp, it's worse than that: We should disallow unaligned sizes.

PR #78, opened 2018-12-12
First tests

Orycterope:
A few tests for Kernel's basic structures, and small fixes to the bugs they helped discover.

### DIFF
--- a/kernel/src/frame_allocator/i386.rs
+++ b/kernel/src/frame_allocator/i386.rs
@@ -169,7 +169,7 @@ impl FrameAllocatorTrait for FrameAllocator {
         assert!(allocator.initialized, "The frame allocator was not initialized");
 
         let mut start_index = 0usize;
-        while start_index + nr_frames < allocator.memory_bitmap.bit_length() {
+        while start_index + nr_frames <= allocator.memory_bitmap.bit_length() {
             let mut temp_len = 0usize;
             loop {
                 match allocator.memory_bitmap.get_bit(start_index + temp_len) {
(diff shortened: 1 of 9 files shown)


roblabla:
> right: `false`: frame_allocator::init() was called twice', kernel/src/frame_allocator/i386.rs:344:5

Didn't you push a fix to that?

Orycterope:
I figured out that the actual problem here is that tests were run in parallel, and this created a run condition where at some point a thread tried to initialize the frame allocator a second time. I tried to reproduce the test fail on my machine to understand how it could have gone wrong, and i ran into the same problem.

For some reason on your server tests were always running serially. 

This is a bug, tests checking the frame allocator should be mutually exclusive. I'm on it.

Orycterope:
We really ought to have tests for our tests 🙃

kernel/src/frame_allocator/i386.rs
```diff
@@ -379,3 +383,309 @@ pub fn mark_frame_bootstrap_allocated(addr: PhysicalAddress) {
...
     allocator.memory_bitmap.set_bit(bit, FRAME_OCCUPIED);
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    const ALL_MEMORY: usize = FRAMES_BITMAP_SIZE * 8 * PAGE_SIZE;
+
+    /// `cargo test` runs tests in parallel, but each test should have its own view of the
```
roblabla: No. Look at `thread_local!` macro.

PR #88, opened 2018-12-14
clippy

### LINKED ISSUE #68: Use clippy on CI to get even more warnings
Because we clearly don't have enough warnings.

Clippy is a cool tool that's able to catch a lot of logic bugs, in addition to imposing some style choices (which helps keep the code uniform).

Orycterope:
Fixes #68 

clippy lint


roblabla:
MFW there's an xclippy command.

So, clippy has something that turns all warnings into errors. See https://github.com/MegatonHammer/linkle/blob/master/ci/script.sh#L16 (-D warnings means "deny warnings")

roblabla:
Closes #68

roblabla:
boop

roblabla:
Missing:

- libuser (nearly half of the errors are from libuser now)
- kernel::ipc
- userspace apps

We're almost there \o/

roblabla:
We're almost there. Only a few warnings in vi are left. I'm going to fix them tomorrow when my head is fresh because they might be tricky and be pointing to actual bugs...

PR #119, opened 2019-01-03
Kernel caps

### LINKED ISSUE #29: Implement kernel access controls (beyond IOPB)
We should implement KACs like Horizon/NX does. We currently have an ad-hoc system for specifying IOPB restrictions, ideally we should merge it with the KAC system.

https://switchbrew.org/w/index.php?title=NPDM#Kernel_Access_Control

roblabla:
Fixes #29 

Horizon/NX is a capability-based OS. It's time we reflect this. Each process now sports a ProcessCapability, which lists everything the process is allowed to do, from syscalls to IRQs. The kernel capability parsing is accurate: errors are returned exactly the way horizon/NX does.

The kernel capabilities in ELF files can be found in the ELF .kernel_caps section. The old .kernel_iopb section is gone: IOPBs are now found in the .kernel_caps section.


roblabla:
@Orycterope Ready for initial review.

PR #133, opened 2019-01-08
libuser: Implement log

roblabla:
Make all libuser programs have a log implementation by default.


libuser/src/log_impl.rs
```diff
@@ -0,0 +1,62 @@
...
+impl Log for Logger {
+    fn enabled(&self, _metadata: &Metadata) -> bool {
+        true
+    }
+
+    fn log(&self, record: &Record) {
+        let _ = writeln!(Logger, "[{}] - {} - {}", record.level(), record.target(), record.args());
+    }
+
+    fn flush(&self) {}
```
Orycterope: flush does nothing, is it normal ?

roblabla: What would it do?

roblabla: Flush should basically ensure all logs have been written to wherever. We already do that because all our logs are backed by writeln, which is guaranteed to be immediately written to output_debug_write (since our backing logger writes on \n). So yeah, flush doing nothing is normal.

PR #157, opened 2019-01-17
Rust2018

### LINKED ISSUE #57: Rust 2018 here we come!
Rust 2018 has so many awesome feature, it'd be a shame if we missed out on them. We should upgrade to rust 1.31, and upgrade all our crates to the 2018 edition.

roblabla:
All I had to do was run cargo fix twice \o/

Fixes #57


roblabla:
🤔 what happened to the todo?

PR #200, opened 2019-03-09
libuser: ZeroBox

Orycterope:
Box that bzero and does not stack overflow.

This module would ideally be in libutils, to be available to the kernel too,
but I didn't want to make libutils depend on alloc.

I considered extending Box<T> when T implements our trait,
instead of exposing some cheap wrapper.
I don't know what would have been better.

### DIFF
--- a/libuser/src/zero_box.rs
+++ b/libuser/src/zero_box.rs
@@ -1,0 +1,75 @@
+//! A module that allocates zeroed types on the heap without copying them from the stack first.
+//!
+//! This is useful for big types that would otherwise cause a stack overflow.
+
+use alloc::boxed::Box;
+use core::ops::{Deref, DerefMut};
+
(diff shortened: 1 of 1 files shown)


roblabla:
I'm not sure what's the point of this? new_zeroed is very unsafe and should be marked as such.

Orycterope:
I'm using it a lot in the ahci driver to allocate big structures on the heap (we're talking multiple pages), which would stack overflow if I were to use Box::new(). 

About the safety, `new_zeroed` is defined only for types that implement our (unsafe) marker trait `ZeroInitialized`, which only serves to convey the meaning that this is safe (see its doc).

Unsafety does not happen at `new_zero`, but when we choose to implement `ZeroInitialized` on a type. From there, I don't see how `new_zero` is unsafe.

PR #201, opened 2019-03-09
Syscalls physical memory

Orycterope:
- syscall MapMmioRegion
- syscall QueryPhysicalAddress
- moving some memory related functions of libuser in their own `mem` module
- improving the syscalls macro
- fixing the capabilities macro


roblabla:
Cargo test fails.

PR #218, opened 2019-03-14
Terminal window size

Orycterope:
Say hello to KFS running in VirtualBox !
Screen Shot 2019-03-14 at 3 44 33 PM

### DIFF
--- a/libuser/src/window.rs
+++ b/libuser/src/window.rs
@@ -50,7 +50,8 @@ pub struct Window {
 
 impl Window {
     /// Creates a window in the vi compositor.
-    pub fn new(vi: &mut ViInterface, top: i32, left: i32, width: u32, height: u32) -> Result<Window, Error> {
+    pub fn new(top: i32, left: i32, width: u32, height: u32) -> Result<Window, Error> {
+        let mut vi = ViInterface::raw_new()?;
         let bpp = 32;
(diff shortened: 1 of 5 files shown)


libuser/src/window.rs
```diff
@@ -50,9 +50,7 @@ pub struct Window {
 
 impl Window {
     /// Creates a window in the vi compositor.
-    pub fn new(top: i32, left: i32, width: u32, height: u32) -> Result<Window, Error> {
-        let mut vi = ViInterface::raw_new()?;
-
+    pub fn new(vi: &mut ViInterface, top: i32, left: i32, width: u32, height: u32) -> Result<Window, Error> {
```
roblabla: Window shouldn't take the ViInterface as parameter. It should reconnect to `vi:`. Eventually, we'll want to have a cache of connections to various sysmodules, such that doing `ViInterface::new()` will returns a shared handle.

https://github.com/MegatonHammer/megaton-hammer/blob/8c183e89d6b60ce592ec91f3b61b9159e1caf187/megaton-hammer/src/ipcdefs/nn/pcie/detail/impl_i_manager.rs#L23-L45

This is how megaton-hammer deals with this.

Orycterope: Oh ok ... i'm changing that right away

Issue #220, opened 2019-03-14
vi: Dropping a window displays corruption


Orycterope:
It should redraw what's behind, but instead this happens : 
![Screen Shot 2019-03-14 at 5 08 32 PM](https://user-images.githubusercontent.com/6343527/54372614-f2351e80-467b-11e9-9636-87148c378ba3.png)

Is this related to #219 ?

Orycterope:
Fixed by #243 apparently

PR #243, opened 2019-03-21
vi: fix window cropping

### LINKED ISSUE #219: vi: fix window cropping
When outside the screen, lines should be cropped, but it's not what happens:
Screen Shot 2019-03-14 at 5 01 51 PM
Window starts at x: 800; y: 0

Orycterope:
Fixes #219


Orycterope:
Also fixes #220 apparently

Issue #216, opened 2019-03-14
userspace.ld stores .bss in the ELF


Orycterope:
Rejoice, it's that "loosing all sanity while screaming at lld" time again !
I know you missed it.

Here's the problem: 

I create a really big array (32M) in vi:
```rust 
// UHD 4K resolution, lives in the .bss
static BACKBUFFER_ARR: Mutex<[VBEColor; (3840 * 2160)]> = Mutex::new([VBEColor::rgb(0, 0, 0); 3840 * 2160]);
```
compile it, make sure that it actually is in the .bss
```shell
$ nm -S kfs-vi | grep BACKBUFFER
00455030 01fa4001 b _ZN6kfs_vi14BACKBUFFER_ARR17he562314c92e24f7eE
```
mmmh 'b' so it's in the .bss, and 0x1fa4001 is the right size. 

Oh Great ! Everything works perf-

```shell
$ ls -lh kfs-vi
-rwxr-xr-x 1 orycterope orycterope  36M Mar 13 23:03 kfs-vi
```

Oh no ...

.... .... how ... i mean ... how ... what ??

Ok maybe roblabla fucked up `userspace.ld`, and it just doesn't put the .bss as NOBITS ?

```shell
$ readelf -S kfs-vi
Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  ...
  [ 6] .data             PROGBITS        00455000 055000 000030 00  WA  0   0  4
  [ 7] .dynamic          NOBITS          00455030 055030 000000 00  WA  0   0  1
  [ 8] .bss              NOBITS          00455030 055030 1fa4148 00  WA  0   0  4
  [ 9] .kernel_caps      PROGBITS        023f9178 1ff9178 000018 00   A  0   0  4
  [10] .kernel_ioports   NOBITS          023f9190 1ff9190 000000 00   A  0   0  1
  [11] .comment          PROGBITS        00000000 1ff9190 000012 01  MS  0   0  1
```

Nope. Wait =_= ... WHAAAAAT ???

Ok so .bss **is** NOBITS, it has a plausible size, but notice how it increments file offsets by exactly its size, where every other NOBITS section does not shift it. This means that .bss is correctly registered as NOBITS, but still stored in the ELF.

This seems to be confirmed by objdump's section hexdump:
```shell
$ objdump -j .data -s kfs-vi

isofiles/boot/kfs-vi:     file format elf32-i386

Contents of section .data:
[...]

PR #247, opened 2019-04-05
Support dynamic relocation in libuser

Thog:
This makes every userland programs position-independent.

This PR also correctly separates every section in its appropriate segment.

### DIFF
--- a/i386-unknown-none-user.json
+++ b/i386-unknown-none-user.json
@@ -11,7 +11,7 @@
   "linker-is-gnu": true,
   "pre-link-args": {
     "ld.lld": [
-      "-Tlink.T"
+      "-Tlinker-scripts/userspace.ld"
     ]
   },
   "executables": true,
--- a/i386-unknown-none-user.json
+++ b/i386-unknown-none-user.json
@@ -1,0 +1,26 @@
+{
+  "llvm-target": "i386-unknown-none",
+  "data-layout": "e-m:e-p:32:32-f64:32:64-f80:32-n8:16:32-S128",
+  "arch": "x86",
+  "target-endian": "little",
+  "target-pointer-width": "32",
+  "target-c-int-width": "32",
+  "os": "none",
+  "linker": "rust-lld",
+  "linker-flavor": "ld.lld",
+  "linker-is-gnu": true,
+  "pre-link-args": {
+    "ld.lld": [
+      "-Tlink.T"
+    ]
+  },
+  "executables": true,
+  "env": "user",
+  "position-independent-executables": true,
+  "dynamic-linking": false,
+  "has-elf-tls": false,
+  "has-rpath": false,
+  "features": "-mmx,-sse,+soft-float",
+  "disable-redzone": true,
+  "panic-strategy": "abort"
+}
--- a/kernel/src/elf_loader.rs
+++ b/kernel/src/elf_loader.rs
@@ -108,6 +108,7 @@ pub fn load_builtin(process_memory: &mut ProcessMemory, module: &MappedGrubModul
 
     // return the entry point
     // TODO: ASLR
+    // BODY: We should generate a random aslr base.
     let entry_point = 0x400000 + elf.header.pt2.entry_point();
     info!("Entry point : {:#x?}", entry_point);
 
@@ -134,7 +135,6 @@ fn load_segment(process_memory: &mut ProcessMemory, segment: ProgramHeader<'_>, 
         flags |= MappingAccessRights::EXECUTABLE
     }
 
-    // TODO: ASLR
     let virtual_addr = 0x400000 + segment.virtual_addr() as usize;
 
     // Create the mapping in UserLand
--- a/kernel/src/elf_loader.rs
+++ b/kernel/src/elf_loader.rs
@@ -107,7 +107,8 @@ pub fn load_builtin(process_memory: &mut ProcessMemory, module: &MappedGrubModul
     }
 
     // return the entry point
-    let entry_point = elf.header.pt2.entry_point();
+    // TODO: ASLR
+    let entry_point = 0x400000 + elf.header.pt2.entry_point();
     info!("Entry point : {:#x?}", entry_point);
 
     entry_point as usize
@@ -133,8 +134,11 @@ fn load_segment(process_memory: &mut ProcessMemory, segment: ProgramHeader<'_>, 
         flags |= MappingAccessRights::EXECUTABLE
     }
 
+    // TODO: ASLR
+    let virtual_addr = 0x400000 + segment.virtual_addr() as usize;
+
     // Create the mapping in UserLand
-    let userspace_addr = VirtualAddress(segment.virtual_addr() as usize);
+    let userspace_addr = VirtualAddress(virtual_addr);
     process_memory.create_regular_mapping(userspace_addr, mem_size_total, flags)
         .expect("Cannot load segment");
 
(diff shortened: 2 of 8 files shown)


roblabla:
I'm curious: Does this fix `vi` being huge because of the .bss being allocated in the file?

Thog:
@roblabla yuuup
image

libuser/src/lib.rs
```diff
@@ -96,24 +99,258 @@ use core::alloc::Layout;
...
+
+/// The runtime linker computes the corresponding virtual address by adding the virtual address at which the shared object is loaded to the relative address.
+const R_386_RELATIVE: usize = 8;
+
+/// Executable entrypoint. Handle relocations and calls real_start.
 #[cfg(target_os = "none")]
+#[naked]
 #[no_mangle]
+#[link_section = ".text.crt0"]
 pub unsafe extern fn start() -> ! {
```
roblabla: This is wrong. A naked function cannot return !, or call `intrinsics::unreachable!()`. A naked function can only have a signature of the form `fn start() -> ()` and may only contain a single ASM block, nothing else.

shell/src/main.rs
```diff
@@ -145,8 +146,8 @@ fn test_threads(terminal: Terminal) -> Terminal {
         unsafe {
             asm!("
             push eax
-            call $0
-            " :: "i"(thread_b as *const u8) :: "intel");
+            call thread_b
```
roblabla: Why was this change made? I suppose "i" is behaving buggy?

Thog: @roblabla It wasn't compiling (I suppose it was trying to hardcode the address of thread_b)

roblabla: Oh, yeah, duh. I wish rust had a "symbol" input type where we could give it a function and it would put the properly mangled symbol. Maybe I should propose this 🤔

libuser/src/lib.rs
Thog: Addressed comment

kernel/src/elf_loader.rs
```diff
@@ -133,8 +134,11 @@ fn load_segment(process_memory: &mut ProcessMemory, segment: ProgramHeader<'_>,
         flags |= MappingAccessRights::EXECUTABLE
     }
 
+    // TODO: ASLR
+    let virtual_addr = 0x400000 + segment.virtual_addr() as usize;
```
Orycterope: This adds an offset even when ASLR is off, and breaks gdb debugging of userspace programs.

The simple way to keep it working is to make the offset 0x00000000 when ASLR is turned off, and link .text to 0x400000 (as it was) for NULL pointers. Also update entry_point offset.

libuser/src/lib.rs
```diff
@@ -96,24 +99,258 @@ use core::alloc::Layout;
...
+            let reloc_type = (*rel).info & 0xff;
+
+            if let R_386_RELATIVE = reloc_type {
+                let ptr = aslr_base.add((*rel).offset) as *mut usize;
+                *ptr += aslr_base as usize;
+            }
+        }
+    }
+
+    // Clean .bss
```
Orycterope: Unfortunately, you must clean the bss **before** you can call any rust (or C) function (that's the whole point of crt0). This means you must do it in asm.

Orycterope:
As I understand (your PR comment wasn't really explicit), this makes kernel builtins illegible for ASLR. But wouldn't relocations for every other userspace programs be handled by the process loader service on behalf of the starting process, instead of having it relocate itself ?

This means the _start will be different for a builtin and a regular process, but they both use libuser. We might want to add some conditional compilation in libuser for that.

i386-unknown-none-user.json
```diff
@@ -0,0 +1,30 @@
+{
+  "llvm-target": "i386-unknown-none",
```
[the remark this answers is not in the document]
roblabla: let's not rename the target right now. Let's do that in a separate PR that does the rename in one fell swoop.

kernel/src/elf_loader.rs
roblabla: > breaks GDB debugging of userspace programs

It shouldn't? This is basically doing "ASLR", with a static random value.

libuser/src/lib.rs
Thog: @Orycterope Not if I ensure that *nothing* use the bss

roblabla: while this is true in principle, all the functions used above (add, offset, size_of) are basically guaranteed not to use the BSS (there's no real compiler guarantee, but it's a sane thing to assume). I think this is fine, and allows using an optimized memset instead of a slow asm copy.

roblabla: Should probably add a comment about it though.

kernel/src/elf_loader.rs
Orycterope: It does breaks GDB, I tried. GDB expects the .text to be at a given address, and blindly puts a breakpoint in virtual memory there. The .text simply isn't there, and nobody told GDB.

Orycterope:
```
(gdb) help add-symbol-file
Load symbols from FILE, assuming FILE has been dynamically loaded.
Usage: add-symbol-file FILE [-readnow | -readnever] [-o OFF] [ADDR] [-s SECT-NAME SECT-ADDR]...
ADDR is the starting address of the file's text.
Each '-s' argument provides a section name and address, and
should be specified if the data and bss segments are not contiguous
with the text.  SECT-NAME is a section name to be loaded at SECT-ADDR.
OFF is an optional offset which is added to the default load addresses
of all sections for which no other address was specified.
```

So I tried
```
target remote :9191
file isofiles/boot/kfs-kernel
add-symbol-file isofiles/boot/kfs-shell 0x00400000
```

And it works smoothly !

Orycterope:
Hum ... Travis still hasn't responded yet, and Mergeable isn't even here. I think they didn't like the repository transfer. @roblabla can you do something about your mergeable bot ?

Orycterope:
Forcing travis rebuild: closing and reopening PR

linker-scripts/userspace.ld
```diff
@@ -1,11 +1,21 @@
 ENTRY(start)
 OUTPUT_FORMAT(elf32-i386)
 
+PHDRS
+{
+  text PT_LOAD FLAGS(5);
+  rodata PT_LOAD FLAGS(4);
+  data PT_LOAD FLAGS(6);
+  bss PT_LOAD FLAGS(6);
```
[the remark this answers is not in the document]
Thog: I prefer to have a clear separation between progbits & nobits sections, that can help to catch some issues :P

kernel/src/elf_loader.rs
Thog: Nice :+1:

i386-unknown-none-user.json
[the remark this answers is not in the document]
Thog: Also @Orycterope "user" is define in "env"

Orycterope:
Closes #216

Thog:
Rebased

kernel/src/elf_loader.rs
```diff
@@ -133,8 +134,11 @@ fn load_segment(process_memory: &mut ProcessMemory, segment: ProgramHeader<'_>,
         flags |= MappingAccessRights::EXECUTABLE
     }
 
+    // TODO: ASLR
```
roblabla: Remove this TODO so we don't get duplicate issues with TODO bot. Also, add a BODY to the above TODO instead :)

Issue #197, opened 2019-03-08
Find a proper name for KFS OS and Kernel


roblabla:
KFS is pretty uninspired. We need something better.

Atmosphere-NX has some really cool names with nice metaphores. The different parts of the OS are named after atmosphere layers (exosphere, thermosphere, stratosphere, troposphere, mesosphere), and the bootloader is a rocket (fusee).

Another example is Fuchsia, which uses precious metals for its component names (zircon, topaz, peridot, garnet).

@Orycterope had the idea to call the OS "Sunrise", after the SynthWave sunrise imagery. ![Sun Rise Synthwave](https://d2tml28x3t0b85.cloudfront.net/tracks/artworks/000/628/933/original/bf22fb.jpeg?1506702399)

Is there more common symbols in the Synthwave culture we could use to name the various components? Any other cool ideas?

Orycterope:
carlOS ... sweet carlos 😌

roblabla:
SunriseOS it is. Now we just need to rename all the old KFS stuff.

PR #249, opened 2019-04-07
Renaming KFS to SunriseOS

Orycterope:
Because what better way to spend a Sunday ?

See #197

The crates are now:

* sunrise-kernel
* sunrise-bootstrap
* sunrise-shell
* sunrise-libuser
* sunrise-clock
* sunrise-sm
* sunrise-vi
* sunrise-ahci
* sunrise-libutils
* sunrise-libkern

Only the README.md and the welcome message at boot use the full "SunriseOS" name.
Every other places use the shorter name "Sunrise" if possible, or "sunrise"
when lowercase is preferred.

For now all components are named after the (short) name of the project, i.e. the kernel
is "the Sunrise kernel", sm is "the Sunrise service manager", etc.


roblabla:
What an absolute madlad xD

.gitignore
```diff
@@ -2,13 +2,13 @@
...
-isofiles/boot/kfs-*
+isofiles/boot/sunrise-*
 os.iso
 DISK.img
 .idea
 .gdbinit
 .gdb_history
 /CMakeLists.txt
-/kfs.iml
+/sunrise.iml
```
roblabla: Should probably just be *.iml

PR #266, opened 2019-04-19
Proper MemoryTypes in bookkeeping

roblabla:
Properly keep track of the MemoryType/MemoryState in the bookkeeping, allowing more accurate svcQueryMemory, an RTLD implementation, and other great goodies.


kernel/src/paging/cross_process.rs
```diff
@@ -16,90 +16,80 @@
...
-    ///     * `offset + len > mapping.length()`.
-    ///     * `offset + len - 1` would overflow.
-    ///     * `len` is 0.
-    pub fn mirror_mapping(mapping: &Mapping, offset: usize, len: usize) -> Result<CrossProcessMapping<'_>, KernelError> {
+    /// * Error if the mapping is not Shared, as only refcounted mappings can be owned.
+    /// * Error if `offset` + `len` > `mapping` length.
+    /// * Error if `offset` + `len` would overflow.
+    // todo: should be offset + (len - 1), but need to check that it wouldn't overflow in our function
+    /// * Error if `len` is 0.
+    pub fn mirror_mapping(mapping: &Mapping, offset: usize, len: usize) -> Result<CrossProcessMapping, KernelError> {
```
Orycterope: The documentation does not specify what is the behavior when adding an offset to a mapping which already holds an offset by itself. 

I believe the offsets should sum, otherwise we might mirror something *before* the mapping, which would be really odd. Also the caller shouldn't have to know about the mapping already being offset, and compute its desired offset accordingly.

kernel/src/paging/cross_process.rs
```diff
@@ -16,90 +16,80 @@
...
     kernel_address: VirtualAddress,
     /// Stores the desired length.
     len: usize,
-    /// The mapping we remapped from.
-    ///
-    /// Note that a `CrossProcessMapping` has the same lifetime as the mapping it remaps.
-    mapping: &'a Mapping,
+    /// The frames this mapping covers.
+    frames: Arc<RwLock<Vec<PhysicalMemRegion>>>
 }
```
Orycterope: Without its lifetime, a `CrossProcessMapping` is now only a special kind of mapping that unmaps from KernelLand when dropped.

You might want to take this more literally, by having the CrossProcessMapping hold a `Mapping` (which itself holds the Arc of the frames), and let Mapping's implementation handle all the offset and len arithmetics, which it already does. That way they're only defined in one place.

kernel/src/interrupts/syscalls.rs
```diff
@@ -137,10 +140,11 @@ fn query_physical_address(virtual_address: usize) -> Result<(usize, usize, usize
     let proc = scheduler::get_current_process();
     let mem = proc.pmemory.lock();
     let mapping = mem.query_memory(virtual_address);
-    let frames = match mapping.mapping().mtype_ref() {
-        MappingType::Regular(regions) => regions,
-        MappingType::Shared(arc_regions) => arc_regions.as_ref(),
-        MappingType::Available | MappingType::Guarded | MappingType::SystemReserved =>
+    let keep_region;
```
Orycterope: `query_physical_address()` does not account for when `mapping.offset != 0` (when the mapping is also offsetted within the frames).

kernel/src/paging/mapping.rs
```diff
@@ -266,164 +269,47 @@ mod test {
...
-        assert_eq!(right.address(), VirtualAddress((5 + 2) * PAGE_SIZE));
-        assert_eq!(right.length(), 1 * PAGE_SIZE);
-        if let MappingType::Regular(held_frames) = right.mtype_ref() {
-            assert_eq!(held_frames.iter().flatten().count(), 1)
-        } else {
-            panic!("test failed, splitting changed type")
-        }
+        let _mapping_err = Mapping::new(VirtualAddress(usize::max_value() - 2 * PAGE_SIZE), MappingFrames::Shared(frames), 0, 2 * PAGE_SIZE, MemoryType::Stack, flags).unwrap_err();
     }
 }
```
Orycterope: Add tests for the offset of a partial mapping. 

I don't believe for a second that we'll get them right from the first time.

kernel/src/paging/process_memory.rs
```diff
@@ -486,24 +449,17 @@ impl ProcessMemory {
...
-                // map them.
-                self.get_hierarchy().map_to_from_iterator(new_frames.iter().flatten(), start_addr + old_size, flags);
-                // create a mapping from the two parts.
-                frames.append(&mut new_frames);
-                Mapping::new_regular(start_addr, frames, flags)
-                    .expect("expand_mapping: couldn't recreate mapping")
-            },
+        let new_mapping = if let MappingFrames::Shared(frames) = old_mapping.frames() {
+            // allocate the new frames.
+            let mut new_frames = FrameAllocator::allocate_frames_fragmented(added_length)?;
```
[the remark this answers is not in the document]
roblabla: Rather than that, removing from bookkeeping should be avoided. Frames are behind a RwLock now.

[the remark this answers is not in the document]
roblabla: Wait no I'm stupid, still need to create a new Mapping for the bookkeeping.

kernel/src/paging/cross_process.rs
roblabla: > Also the caller shouldn't have to know about the mapping already being offset, and compute its desired offset accordingly.

That was the intent, but obviously, I failed.

I'm going to refactor this code to reuse the `Mapping::new` logic. I'll still need to have a bit of logic to handle the low bits of the offset. A CrossProcessMapping can have an offset in the middle of the page.

kernel/src/paging/mapping.rs
```diff
@@ -112,6 +117,48 @@ impl Mapping {
     /// Returns the frames in this mapping.
     pub fn frames(&self) -> &MappingFrames { &self.frames }
 
+    /// Returns an iterator over the Physical Addresses mapped by this region.
+    /// This takes into account the physical offset and the length of the
+    /// mapping.
+    pub fn frames_it(&self) -> impl Iterator<Item = PhysicalAddress> + Clone + core::fmt::Debug + '_ {
+        /// Anonymous iterator over mapping frames' PhysicalAddresses.
+        #[derive(Debug, Clone)]
+        struct MappingFramesIt<'a>(&'a MappingFrames, usize, StepBy<Range<usize>>);
```
Orycterope: The iterator should not lock and unlock the frames as it advances. It would make no sense to enable someone to modify the frames while they are being iterated. 

The lock should be held for the whole lifetime of the iterator.

kernel/src/paging/mapping.rs
```diff
@@ -112,6 +117,48 @@ impl Mapping {
     /// Returns the frames in this mapping.
     pub fn frames(&self) -> &MappingFrames { &self.frames }
 
+    /// Returns an iterator over the Physical Addresses mapped by this region.
+    /// This takes into account the physical offset and the length of the
+    /// mapping.
+    pub fn frames_it(&self) -> impl Iterator<Item = PhysicalAddress> + Clone + core::fmt::Debug + '_ {
```
Orycterope: Great ! You should use it in process_mapping when giving a mapping to the page tables.

kernel/src/paging/mapping.rs
Orycterope: Your implementation is extremely complicated for something as simple as 

```
mapping.read().iter().flatten()
    .skip(phys_offset / PAGE_SIZE)
    .take(length / PAGE_SIZE)
```

I know you also have to keep the lock hanging somewhere, but common, what is this iterative StepBy nonsense ?

roblabla: > `mapping.read()`
Have to keep the lock
> `.iter()`
Have to abstract over Owned/Shared/None
> `.flatten()`
Can't use flatten because then I'd need to expose the PhysicalMemRegions, which I can't without doing even more locking shenanigans

PR #270, opened 2019-04-22
PCI: Move to libuser. Implement 64-bit BARs.

roblabla:
Share the PCI implementation across multiple drivers by moving it to libuser. Should eventually move to its own service, but this is a good temporary measure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Orycterope:
news on this PR ? Is this ready for review ?

roblabla:
I'll probably have to re-do it entirely.

Orycterope:
good call

PR #306, opened 2019-05-23
Better logging - more colors, less duplication

roblabla:
Brings three improvements:

- Puts some color on the log level, which is neat
- Allows filtering of logs coming from userspace applications in grub. We can now do `sunrise_clock=debug,info` to get more debug information from sunrise_clock, for instance.
- Makes the outputDebugString aware of the log formatting, which results in logs that look like the following:

```
[INFO] - sunrise_clock - clock - 13:59:48 Thursday 23 May 19
```

instead of the old:

```
[INFO] - sunrise_kernel::interrupts::syscalls - clock - [INFO] - sunrise_clock - 13:59:48 Thursday 23 May 19
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -502,6 +502,7 @@ dependencies = [
 name = "sunrise-clock"
 version = "0.1.0"
 dependencies = [
+ "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "sunrise-libuser 0.1.0",
 ]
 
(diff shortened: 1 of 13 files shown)


Orycterope:
And if by any chance you figure out why colors end up fucking up the terminal it would be pretty great.

roblabla:
> However, I don't see the code that handles filtering from grub arguments.

Weeelp, forgot to commit that bit. Will come in a minute. When it'll be implemented, it'll go through the args of the kernel. Basically, `multiboot2 /boot/sunrise-bootstrap "sunrise_clock=debug,info"` will do what you'd expect.

PR #308, opened 2019-05-30
Update to nightly-2019-05-22

roblabla:
- Moved all the git dependencies we kept in our private repos to SunriseOS/.
- Alloc, alloc, alloc


roblabla:
Here's to hoping this broke nothing.

PR #311, opened 2019-06-02
Replace object macro with a saner trait-based solution

roblabla:
The object macro was a bit of a crazy attempt to port the (already insane) C++ template-based IPC dispatcher found in libstratosphere to Rust. It served its purpose well, but it was hard to maintain, and made every code contained in it hard to debug. It was a really sub-par experience.

In this PR, I attempt to replace it with an idea I already discussed a few times: Generate traits for every interface, and "implement Object" on all those traits. swipc-gen will now generate two items for every interface:

- a `trait Interface` that are used for building servers.
- A `struct InterfaceProxy`, which is the same old client interface we used to
  have.

However, there is a bit of a problem with "implementing Object" on every interface: coherence rules prohibits multiple blanket implementations ([here's a link to the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=df8d76ae5f72eadf9dc8c70d59356f54)). After a bit of experimentation, I settled on a fairly simple pattern: Give PortHandler/SessionWrapper a function pointer to the dispatch function instead of going through Object. This dispatch function will get generated as a default function in the trait interface.

Hopefully, this new system will be a lot saner to deal with. It should also be much easier to plug Futures into this, among other things.


swipc-gen/src/gen_rust_code.rs
```diff
@@ -445,6 +460,277 @@ fn generate_mod(m: Mod, depth: usize, mod_name: &str, crate_name: &str) -> Strin
...
+            writeln!(s, "            {} => {{", func.num).unwrap();
+            writeln!(s, "{}", val).unwrap();
+            writeln!(s, "            }},").unwrap();
+        } else {
+            writeln!(s, "            // Unsupported: {}", func.num).unwrap();
+        }
+    }
+    writeln!(s, "            _ => {{").unwrap();
+    writeln!(s, "                let mut msg__ = Message::<(), [_; 0], [_; 0], [_; 0]>::new_response(None);").unwrap();
+    writeln!(s, "                msg__.set_error(sunrise_libkern::error::KernelError::PortRemoteDead.make_ret() as u32);").unwrap();
```
Orycterope: unrecognized cmdid returns PortRemoteDead error (which is a kernel error), and returns Ok(). Is this the intended behavior ?

Wouldn't it want to actually drop the session instead ?

roblabla: This is the wanted behavior, yes. Nintendo's to blame here :')

roblabla:
will rebase and merge tomorrow.

PR #315, opened 2019-06-08
Fix paging leak

Orycterope:
See commits

### DIFF
--- a/kernel/src/paging/hierarchical_table.rs
+++ b/kernel/src/paging/hierarchical_table.rs
@@ -202,11 +202,13 @@ pub trait PagingCacheFlusher {
 ///
 /// When passing this struct the TLB will **not** be flushed. Used by Inactive/PagingOff page tables,
 /// and DynamicHierarchy
+#[derive(Debug)]
 pub struct NoFlush;
 impl PagingCacheFlusher for NoFlush { fn flush_whole_cache() { /* do nothing */ } }
 
 /// This is just a wrapper for a pointer to a table.
 /// It enables us to do handle when it is dropped
+#[allow(missing_debug_implementations)]
 pub struct SmartHierarchicalTable<'a, T: HierarchicalTable>(*mut T, PhantomData<&'a T>);
 
 impl<'a, T: HierarchicalTable> SmartHierarchicalTable<'a, T> {
@@ -633,5 +635,26 @@ pub trait InactiveHierarchyTrait : TableHierarchy {
     fn is_currently_active(&self) -> bool;
 
     /// Returns the currently active hierarchy as an inactive hierarchy.
+    ///
+    /// Used only when becoming the first process to get a hold on the page tables
+    /// created by the bootstrap before us, so we can free them.
+    ///
+    /// Dropping it will **not free the pages** owned by this InactiveHierarchy.
+    /// This is fine, because they used to belong to the bootstrap, and are already
+    /// considered free by the [FrameAllocator], so we must leak them.
+    ///
+    /// However, it **will free the tables** (including directory) of this InactiveHierarchy,
+    /// except the ones mapping KernelLand memory, as for any other regular process.
+    /// This frames were marked as occupied when initialising the `FrameAllocator`,
+    /// we're making them available again.
+    ///
+    /// # Unsafety
+    ///
+    /// Having multiple InactiveHierarchy pointing to the same table hierarchy is unsafe.
+    /// Should not be use for any other purpose, it is only guaranteed to be safe to drop.
+    ///
+    /// Make sure you switch to a new table hierarchy before dropping it.
+    ///
+    /// [FrameAllocator]: crate::frame_allocator::FrameAllocator
     unsafe fn from_currently_active() -> Self;
 }
--- a/kernel/src/paging/hierarchical_table.rs
+++ b/kernel/src/paging/hierarchical_table.rs
@@ -631,4 +631,7 @@ pub trait InactiveHierarchyTrait : TableHierarchy {
     /// Generally this means comparing the current MMU register pointer to top-level table with the
     /// address of the top-level table of this hierarchy.
     fn is_currently_active(&self) -> bool;
+
+    /// Returns the currently active hierarchy as an inactive hierarchy.
+    unsafe fn from_currently_active() -> Self;
 }
--- a/kernel/src/paging/hierarchical_table.rs
+++ b/kernel/src/paging/hierarchical_table.rs
@@ -2,11 +2,9 @@
 
 // what the architecture code still has define
 use super::arch::{PAGE_SIZE, ENTRY_COUNT};
-use super::lands::{RecursiveTablesLand, VirtualSpaceLand};
 use super::MappingAccessRights;
 
 use crate::mem::{VirtualAddress, PhysicalAddress};
-use crate::frame_allocator::{PhysicalMemRegion};
 use crate::utils::align_up_checked;
 use core::marker::PhantomData;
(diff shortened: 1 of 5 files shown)


kernel/src/paging/hierarchical_table.rs
```diff
@@ -596,37 +594,33 @@ pub trait TableHierarchy {
...
 }
 
-/// A trait implemented by innactive table hierarchies.
-/// Enables creating a
+/// A trait implemented by inactive table hierarchies.
+///
+/// Extends the [TableHierarchy] trait by adding functions to switch to this hierarchy,
+/// when process-switching.
+///
+/// # Drop
```
roblabla: Why is Drop behavior explained on the Trait? Shouldn't this be on the i386 implementation struct ?

kernel/src/process.rs
```diff
@@ -440,8 +440,16 @@ impl ProcessStruct {
     /// Panics if max PID has been reached, which it shouldn't have since we're the first process.
     unsafe fn create_first_process() -> Arc<ProcessStruct> {
 
-        // the already currently active pages
-        let pmemory = Mutex::new(ProcessMemory::from_active_page_tables());
+        let mut pmemory = ProcessMemory::default();
+        pmemory.switch_to();
+
+        // * just leaked bootstrap page table hierarchy right here *
```
roblabla: Couldn't we trivially free them by creating an InactiveHierarchy from the current cr3? E.G., do this:

```rust
        let mut pmemory = ProcessMemory::default();
        let bootstrap_cr3 = super::read_cr3();
        pmemory.switch_to(); 

        // Destroy bootstrap page tables
        core::mem::drop(InactiveHierarchy { directory_physical_address: PhysicalAddress(bootstrap_cr3) });
```

What's wrong with such a naive approach ?

kernel/src/paging/hierarchical_table.rs
Orycterope: Actually it's documented on both. It's part of the contract that InactiveHierarchyTrait creates, or at least very relative to that contract, and gives a good hint about how an implementer should handle the drop, so I think it's appropriate to document it here.

kernel/src/process.rs
Orycterope: We could, and it would work, however InactiveHierarchy is not exposed, it is private to the paging module, and I would rather keep it that way.

The only structure we can create here is a full ProcessMemory for the bootstrap, which itself holds this InactiveHierarchy. This is the solution I came up a few days ago, the diff looked like this:

```
diff --git a/kernel/src/paging/process_memory.rs b/kernel/src/paging/process_memory.rs
index 88400998..5ace9e7d 100644
--- a/kernel/src/paging/process_memory.rs
+++ b/kernel/src/paging/process_memory.rs
@@ -164,11 +164,27 @@ impl Default for ProcessMemory {
 
 impl ProcessMemory {
     /// Creates a ProcessMemory referencing the current page tables.
-    /// Used only when becoming the first process for creating the first ProcessMemory.
+    ///
+    /// Used only when becoming the first process to get a hold on the page tables
+    /// created by the bootstrap before us, so we can free them.
+    ///
+    /// The ProcessMemory will have an empty [UserspaceBookkeeping]. This means that
+    /// dropping it will **not free the pages** owned by this ProcessMemory.
+    /// This is fine, because they used to belong to the bootstrap, and are already
+    /// considered free by the [FrameAllocator], so we must leak them.
+    ///
+    /// However, it **will free the tables** (including directory) of this ProcessMemory,
+    /// except the ones mapping KernelLand memory, as for any other regular process.
+    /// This frames were marked as occupied when initialising the `FrameAllocator`,
+    /// we're making them available again.
     ///
     /// # Unsafety
     ///
-    /// Having multiple ProcessMemory pointing to the same table hierarchy is unsafe.
+    /// Should not be use for any other purpose, most of the values in the returned ProcessMemory
+    /// are mocked. It is only guaranteed to be safe to drop.
+    ///
+    /// Should only be called once.
+    /// Having multiple ProcessMemory pointing to the same table hierarchy is undefined behavior.
     pub unsafe fn from_active_page_tables() -> Self {
         ProcessMemory {
             userspace_bookkeping: UserspaceBookkeeping::new(),
             table_hierarchy: InactiveHierarchy::from_currently_active(),
             heap_base_address: VirtualAddress(0x55555555), // just a dummy value, the first process
                                                            // should never use its process's heap !
         }

diff --git a/kernel/src/process.rs b/kernel/src/process.rs
index d67a28ec..47dc9961 100644
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -440,8 +440,15 @@ impl ProcessStruct {
     /// Panics if max PID has been reached, which it shouldn't have since we're the first process.
     unsafe fn create_first_process() -> Arc<ProcessStruct> {
 
-        // the already currently active pages
-        let pmemory = Mutex::new(ProcessMemory::from_active_page_tables());
+        // get the page hierarchy currently in use that was set-up by the bootstrap, so we can free it.
+        let bootstrap_pages = ProcessMemory::from_active_page_tables();
+
+        // allocate new memory space
+        let mut pmemory = ProcessMemory::default();
+        // switch to it
+        pmemory.switch_to();
+        // and free the old bootstrap pages
+        drop(bootstrap_pages);
 
         let pid = NEXT_PROCESS_ID.fetch_add(1, Ordering::SeqCst);
         if pid == usize::max_value() {
@@ -452,7 +459,7 @@ impl ProcessStruct {
             ProcessStruct {
                 pid,
                 name: String::from("init"),
-                pmemory,
+                pmemory: Mutex::new(pmemory),
                 threads: SpinLockIRQ::new(Vec::new()),
                 phandles: SpinLockIRQ::new(HandleTable::default()),
                 killed: AtomicBool::new(false),
```

At that point the really weird contracts it created, and the few unsafe and completely backward api functions it requires did not seemed worth the trouble for saving just 2 frames. I'd rather have a much simpler and purer api for the paging, and accept to sacrifice 2 frames. 

If in the end you still disagree, just tell me, i'll revert my commit and apply this patch instead.

kernel/src/paging/hierarchical_table.rs
```diff
@@ -639,5 +635,26 @@ pub trait InactiveHierarchyTrait : TableHierarchy {
...
+    /// Used only when becoming the first process to get a hold on the page tables
+    /// created by the bootstrap before us, so we can free them.
+    ///
+    /// Dropping it will **not free the pages** owned by this InactiveHierarchy.
+    /// This is fine, because they used to belong to the bootstrap, and are already
+    /// considered free by the [FrameAllocator], so we must leak them.
+    ///
+    /// However, it **will free the tables** (including directory) of this InactiveHierarchy,
+    /// except the ones mapping KernelLand memory, as for any other regular process.
+    /// This frames were marked as occupied when initialising the `FrameAllocator`,
```
roblabla:
```suggestion
    /// These frames were marked as occupied when initialising the `FrameAllocator`,
```

kernel/src/paging/hierarchical_table.rs
```diff
@@ -639,5 +635,26 @@ pub trait InactiveHierarchyTrait : TableHierarchy {
...
+    /// we're making them available again.
+    ///
+    /// # Unsafety
+    ///
+    /// Having multiple InactiveHierarchy pointing to the same table hierarchy is unsafe.
+    /// Should not be use for any other purpose, it is only guaranteed to be safe to drop.
```
roblabla:
```suggestion
    /// Should not be used for any other purpose, it is only guaranteed to be safe to drop.
```

PR #317, opened 2019-06-10
 Properly implement A/B/W buffers

Thog:
Properly implement A/B/W buffers in kernel and implement B buffers in libuser.
Also, fix handle<copy> in object! macro.

This is needed by the upcoming time service PR.

### DIFF
--- a/kernel/src/ipc/session.rs
+++ b/kernel/src/ipc/session.rs
@@ -735,17 +735,9 @@ fn pass_message(from_buf: &[u8], from_proc: Arc<ThreadStruct>, to_buf: &mut [u8]
             return Err(UserspaceError::PortRemoteDead)
         }
 
-        let mut current_memlock = if is_reply {
-            from_proc.process.pmemory.lock()
-        } else {
-            to_proc.process.pmemory.lock()
-        };
+        let mut current_memlock = to_proc.process.pmemory.lock();
 
-        let (mut from_mem, mut to_mem) = if is_reply {
-            (&mut *current_memlock, &mut *other_memlock)
-        } else {
-            (&mut *other_memlock, &mut *current_memlock)
-        };
+        let (mut from_mem, mut to_mem) = (&mut *other_memlock, &mut *current_memlock);
 
         for i in 0..hdr.num_a_descriptors() {
             buf_map(from_buf, to_buf, &mut curoff, &mut *from_mem, &mut *to_mem, MappingAccessRights::empty(), buffers)?;
@@ -761,11 +753,7 @@ fn pass_message(from_buf: &[u8], from_proc: Arc<ThreadStruct>, to_buf: &mut [u8]
     }
 
     if is_reply && !buffers.is_empty() {
-        let (mut from_mem, mut to_mem) = if is_reply {
-            (from_proc.process.pmemory.lock(), other_memlock)
-        } else {
-            (other_memlock, to_proc.process.pmemory.lock())
-        };
+        let (mut from_mem, mut to_mem) = (from_proc.process.pmemory.lock(), other_memlock);
 
         // Unmap A-B-W buffers
         for buffer in buffers {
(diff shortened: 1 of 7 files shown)


kernel/src/ipc/session.rs
```diff
@@ -243,12 +269,61 @@ fn buf_map(from_buf: &[u8], to_buf: &mut [u8], curoff: &mut usize, from_mem: &mu
...
+            let mut to = UserSpacePtrMut::from_raw_parts_mut(to_addr.addr() as *mut u8, first_page_size);
+            to.copy_from_slice(&from);
+            size_handled += first_page_size;
+        }
+
+        if (addr + size) % PAGE_SIZE != 0 && (to_addr + size).floor() != to_addr_full {
+            // memcpy the last page.
+            let last_page_size = (addr + size) % PAGE_SIZE;
+
+            let from_mapping = from_mem.mirror_mapping((VirtualAddress(addr) + size).floor(), last_page_size)?;
```
roblabla: If this fails, we should unmap the previously created mirror mapping. (This comment applies to all other error conditions here. I think a good way to do this would be to create a very lightweight "Guard" structure that, when dropped, unmaps, and have the success condition forget the guard. Maybe there's a good crate implementing this?)

kernel/src/ipc/session.rs
```diff
@@ -262,10 +337,66 @@ fn buf_map(from_buf: &[u8], to_buf: &mut [u8], curoff: &mut usize, from_mem: &mu
...
+    let to_addr_full = to_addr.floor();
+    let mut size_handled = 0;
+
+    if addr.addr() % PAGE_SIZE != 0 {
+        let first_page_size = core::cmp::min(PAGE_SIZE - (addr.addr() % PAGE_SIZE), size);
+
+        if buffer.writable {
+            // memcpy the first page.
+            let from = UserSpacePtr::from_raw_parts(addr.addr() as *const u8, first_page_size);
+            let to_mapping = to_mem.mirror_mapping(to_addr, first_page_size)?;
```
roblabla: If this fails, we should still unmap from `from_mem`. So again, we need better error handling here.

kernel/src/ipc/session.rs
```diff
@@ -596,19 +728,48 @@ fn pass_message(from_buf: &[u8], from_proc: Arc<ThreadStruct>, to_buf: &mut [u8]
...
-        let mut from_mem = from_proc.process.pmemory.lock();
-        let mut to_mem = to_proc.process.pmemory.lock();
+        if is_reply {
+            // TODO: Error to return when replying with A/B/W descriptors
+            // BODY: We currently reply with a broken ass error when replying
+            // BODY: with A/B/W descriptors (which is illegal).
+            return Err(UserspaceError::PortRemoteDead)
+        }
+
+        let mut current_memlock = if is_reply {
```
roblabla: `is_reply` can't happen here :eyes:

kernel/src/ipc/session.rs
```diff
@@ -596,19 +728,48 @@ fn pass_message(from_buf: &[u8], from_proc: Arc<ThreadStruct>, to_buf: &mut [u8]
...
         }
 
         for i in 0..hdr.num_w_descriptors() {
-            buf_map(from_buf, to_buf, &mut curoff, &mut *from_mem, &mut *to_mem, MappingAccessRights::WRITABLE)?;
+            buf_map(from_buf, to_buf, &mut curoff, &mut *from_mem, &mut *to_mem, MappingAccessRights::WRITABLE, buffers)?;
+        }
+    }
+
+    if is_reply && !buffers.is_empty() {
+        let (mut from_mem, mut to_mem) = if is_reply {
```
roblabla: !is_reply can't happen here

kernel/src/ipc/session.rs
roblabla:
```rust
match to_mem.mirror_mapping() {
    Ok(to_mapping) => { /* do stuff */ },
    Err(err) => error = err,
}
```

roblabla:
```
// This needs explicit error handling since the user might unmap `to_addr` in-between sending the request and receiving the response
```

PR #318, opened 2019-06-10
Implement a time service

Thog:
No one asked for it but here it is: The Civil Time PR!

This PR adds a new process "time" in charge of handling the RTC and Timezones.
Also, The "clock" process was renamed to "wall-clock" and was updated to use the IPC interface accordingly.  

PS: This PR needs #317


ipcdefs/time.id
```diff
@@ -0,0 +1,83 @@
+# Represent a time zone rule
```
roblabla: Maybe add some documentation explaining that this is an opaque blob of data?

ipcdefs/time.id
```diff
@@ -0,0 +1,83 @@
...
+# Represent a time zone location name
+type sunrise_libuser::time::LocationName = bytes<0x24>;
+
+# Represent a POSIX Timestamp
+type sunrise_libuser::time::PosixTime = u64;
+
+# Represent the basic informations of a local time.
+type sunrise_libuser::time::CalendarTime = struct<0x8> {
+    # The year of the local time
+	i16 year;
```
roblabla: Struct fields should be indented.

libtimezone/src/conversion.rs
```diff
@@ -0,0 +1,977 @@
...
+
+        // Is it in range?
+        if abbreviation_list_index >= timezone_rule.charcnt as u8 {
+            return Err(TimeZoneError::InvalidData);
+        }
+
+        timezone_rule.ttis[i].abbreviation_list_index = i32::from(abbreviation_list_index);
+    }
+
+    // FIXME: clone_fron_slice doesn't do the same thing as this loop. We need to report that on rustc repo.
```
roblabla: IIRC, this is because of padding bytes. TLDR: copy_from_slice/this for loop does a memcpy, whereas clone_from_slices loops over the values. Rust probably transforms multiple memcpys into a big memcpy in both the for loop/copy_from_slice case, but not in the clone_from_slice. So, technically, this is normal.

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,693 @@
+//! Local Time crate
+//!
+//! Permit to manipulate timezone data.
```
roblabla: "Allows manipulating timezone data". 

A (very simple) example usage would be nice?

libtimezone/src/utils.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    if let Some(value) = res {
+        *ip = value;
+        false
+    } else {
+        true
+    }
+}
+
+/// Normalize and increment the given ``ip`` with the given ``unit`` and ``base`` if it doesn't overflow.
```
roblabla: What does "normalize" mean here?

ipcdefs/time.id
```diff
@@ -0,0 +1,92 @@
...
+	# True if the local time represent a Day Saving Time
+	bool is_daylight_saving_time;
+
+	# The GMT offset of the timezone used to generate this local time
+	i32 gmt_offset;
+};
+
+# Entry point interface
+interface sunrise_libuser::time::StaticService is time:u, time:a, time:s {
+	# Get the TimeZone service object
```
roblabla: Documentation should have the same indent level as their command.

ipcdefs/time.id
```diff
@@ -0,0 +1,92 @@
...
+    [3] get_timezone_service() -> object<sunrise_libuser::time::TimeZoneService>;
+}
+
+# RTC interface
+interface sunrise_libuser::time::RTCManager is rtc {
+	# Get the current RTC time
+	[1] get_rtc_time() -> sunrise_libuser::time::PosixTime;
```
roblabla: More indentation. Plz.

libtimezone/src/conversion.rs
```diff
@@ -0,0 +1,977 @@
...
+    res
+}
+
+/// Parse a number and return the rest if the number is in range.
+fn get_num<'a>(name: &'a [u8], num: &mut i64, min: i64, max: i64) -> Option<&'a [u8]> {
+    if !(name[0] as char).is_digit(10) {
+        return None;
+    }
+
+    let name_str = unsafe { core::str::from_utf8_unchecked(name) };
```
roblabla: This makes me very uncomfortable. Isn't `name` user-controlled? If so, this can blow up really badly if the user sends invalid UTF-8. Maybe name should be converted to UTF-8 once early in the code, with `from_utf8`, bailing if the name is not valid utf-8.

libtimezone/src/conversion.rs
```diff
@@ -0,0 +1,977 @@
...
+    timezone_rule: &mut TimeZoneRule,
+    conversion_buffer: &mut ConversionBuffer,
+) -> TimeZoneResult<()> {
+    if conversion_buffer.work_buffer.len() < core::mem::size_of::<TzifHeader>() {
+        return Err(TimeZoneError::InvalidSize);
+    }
+
+    let work_buffer = conversion_buffer.work_buffer;
+
+    let tzheader: &TzifHeader = unsafe { &*(work_buffer as *const _ as *const TzifHeader) };
```
roblabla: Something to be really careful about: tzheader's lifetime isn't tied to the work_buffer's here. You might want to put this in a function that ties the lifetime:

```rust
unsafe fn tzif_header_from_bytes<'a>(work_buffer: &'a [u8]) -> &'a TzifHeader {
    &*(work_buffer as *const _ as *const TzifHeader)
}
```

roblabla: Alternatively, could depend on `plain`, which is a crate for doing this kind of thing safely. `plain` also checks the size (which you do) and the alignment (which you don't).

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,694 @@
...
+#[derive(Copy, Clone, Default, Debug)]
+pub struct CalendarTime {
+    /// The local time basic informations.
+    pub time: CalendarTimeInfo,
+
+    /// Additional information of the local time.
+    pub additional_info: CalendarAdditionalInfo,
+}
+
+//assert_eq_size!(calendartime; CalendarTime, [u8; 0x20]);
```
roblabla: Forgot to re-enable this static assertion?

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,694 @@
...
+                }
+            }
+
+            self.types[lo as usize - 1] as usize
+        };
+
+        let tti = &self.ttis[tti_index];
+        let mut result = create_calendar_time(time, tti.gmt_offset)?;
+        result.additional_info.is_dst = tti.is_dst;
+        // FIXME: copy timezone name
```
roblabla: FIXMEs aren't caught by the TODO system. Rename to TODO, and write a body explaining what's missing? (Unless this is meant to be fixed before merge).

(This applies to all the FIXMEs)

libutils/src/lib.rs
```diff
@@ -274,3 +274,17 @@ mod test {
         log2_ceil(0);
     }
 }
+
+#[macro_export]
+/// A macro to initialize a struct directly in global.
```
roblabla:
```
///
/// # Usage
///
/// ```rust
/// let zero_initialized = unsafe { initialize_to_zero!(u32) };
/// ```

File: time/src/timezone.rs [hunk 13]
```diff
@@ -0,0 +1,326 @@
...
+pub static TZ_MANAGER: Mutex<TimeZoneManager> = Mutex::new(unsafe { initialize_to_zero!(TimeZoneManager) });
+
+/// TimeZone service object.
+#[derive(Default, Debug)]
+pub struct TimeZoneService {
+    /// A dummy field present to just avoid having a zero sized type.
+    pub dummy: u64
+}
+
+impl Drop for TimeZoneService {
```
roblabla: Shouldn't this get removed?

time/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// Checks if the RTC is in 12 hours or 24 hours mode. Depending on the mode,
+    /// the date might be encoded in BCD.
+    #[allow(clippy::wrong_self_convention)] // More readable this way.
+    pub fn is_12hr_clock(&mut self) -> bool {
+        self.read_reg(0xB) & (1 << 2) != 0
+    }
+}
+
+/// Global instance of Rtc. It's safe to actually access it as it isn't modified concurently
+static mut RTC_INSTANCE: Rtc = unsafe { initialize_to_zero!(Rtc) };
```
roblabla: TODO: thread_local, basically. Then we'll be able to put it behind a RefCell<>.

time/src/timezone.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.location
+    }
+
+    /// Set the time zone name used on this devie.
+    /// 
+    /// Note:
+    /// 
+    /// This also load the new timezone rule.
+    pub fn set_device_location_name(&mut self, location: LocationNameInternal) -> IpcResult<()> {
+        let path = format!("zoneinfo/{}", unsafe { core::str::from_utf8_unchecked(&location) });
```
roblabla: 100% unsafe. This *is* a security vulnerability (invalid UTF8 in an &str can cause memory unsafety). Use from_utf8, return an error if it fails.

time/src/timezone.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Get the total count of location name available
+    pub fn get_total_location_name_count(&self) -> IpcResult<(u32, )> {
+        // FIXME: use binaryList.txt
+        Ok((TimeZoneFileSystem::file_count(), ))
+    }
+
+    /// Load a time zone rule.
+    pub fn load_timezone_rule(&mut self, location: LocationNameInternal, timezone_rule: Option<&mut TimeZoneRule>) -> IpcResult<()> {
+        let path = format!("zoneinfo/{}", unsafe { core::str::from_utf8_unchecked(&location) });
```
roblabla: idem

wall-clock/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+
+fn main() {
+    let mut logger = Terminal::new(WindowSize::FontLines(1, true)).unwrap();
+    let mut time = StaticService::raw_new_time_u().unwrap();
+    let mut rtc = RTCManager::raw_new().unwrap();
+    let mut timezone_service = time.get_timezone_service().unwrap();
+
+    // Get default timezone name
+    let device_location = timezone_service.get_device_location_name().unwrap();
+    let device_location_trimed = unsafe { core::str::from_utf8_unchecked(&device_location) }.trim_matches(char::from(0));
```
roblabla: Again, 100% unsafe.

wall-clock/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    // Let's get New York time
+    let custom_location = b"America/New_York\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
+    let custom_location_trimed = unsafe { core::str::from_utf8_unchecked(&custom_location[..]) }.trim_matches(char::from(0));
```
roblabla: TBH, while this one is technically OK, we should just use the safe variant here, nothing will be lost.

libtimezone/src/lib.rs
Thog: urgh totally missed to fix this one...

time/src/timezone.rs
roblabla: In this instance, if location is not utf8, we should return Not Found.

time/src/timezone.rs
roblabla: In this instance, if location is not utf8, we should return Not Found.

wall-clock/src/main.rs
roblabla: So if I understand correctly, this is only needed to print the timezone name. Fenrir suggested bstr, and I think that would work pretty well in this case. The `Display` implementation of `BStr` replaces invalid UTF-8 bytes with the Unicode replacement codepoint (�).

libtimezone/src/conversion.rs
Thog: Well the data come from the timezone archive and isn't really user controlled but you are right I shouldn't be using the unchecked variant

libtimezone/src/lib.rs
Thog: Yep and I should put a comment here.

I have changed the type of the year field of CalendarTimeInfo because of possible wrapping issues during conversion

PS: the type on the IPC side is correct

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+            hour: 0,
+            minute: 0,
+            second: 0,
+        }
+    }
+}
+
+assert_eq_size!(CalendarTimeInfo, [u8; 0x10]);
+
+/// Represnet aditional info of a local time.
```
Orycterope: Bless you

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,704 @@
...
+
+    calendar_time.time.day = (day_of_year + 1) as i8;
+    calendar_time.additional_info.is_dst = false;
+    calendar_time.additional_info.gmt_offset = gmt_offset;
+
+    Ok(calendar_time)
+}
+
+/// Compare two CalendarTimeInfo and return the difference of two of them.
+fn compare_calendar_info(a: &CalendarTimeInfo, b: &CalendarTimeInfo) -> isize {
```
Orycterope: You might impl `Ord` or `PartialOrd` instead

Orycterope: You are comparing local times, this is valid only if they are in the same local. This should be documented.

time/src/timezone.rs
```diff
@@ -0,0 +1,328 @@
...
+        fn load_timezone_rule(&mut self, location: LocationNameInternal, tz_rules: OutBuffer<TimeZoneRule>, ) -> Result<(), Error> {
+            info!("HELLO YOU");
+
+            let mut tz_rules = tz_rules;
+            TZ_MANAGER.lock().load_timezone_rule(location, Some(&mut tz_rules))
+        }
+
+        #[cmdid(5)]
+        #[inline(never)]
+        fn test(&mut self, test: OutBuffer<LocationNameInternal>, ) -> Result<(), Error> {
```
Orycterope: remove me

time/src/main.rs
Orycterope: Soon™

libutils/src/lib.rs
```diff
@@ -274,3 +274,23 @@ mod test {
...
+///
+/// # Usage
+///
+/// ```rust
+/// let zero_initialized = unsafe { initialize_to_zero!(u32) };
+/// ``
```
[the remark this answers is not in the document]
roblabla: Specifically: The type should not contain anything that is not allowed to be initialized to Zero (references, certain enums, and complex types).

I'm not 100% sure if padding bytes are a problem here. I'm pretty sure they aren't.

libtimezone/src/conversion.rs
```diff
@@ -0,0 +1,977 @@
...
+
+    /// Invalid type.
+    Invalid,
+}
+
+#[repr(C, align(8))]
+/// Represent a rule of a POSIX TimeZone name.
+struct Rule {
+    /// The type of this rule.
+    pub rule_type: RuleType,
```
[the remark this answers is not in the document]
Thog: See usage in ``get_rule``

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,703 @@
...
+    /// The count of time transitions.
+    pub(crate) timecnt: i32,
+
+    /// The count of time type infos.
+    pub(crate) typecnt: i32,
+
+    /// The count of chars.
+    pub(crate) charcnt: i32,
+
+    /// ?
```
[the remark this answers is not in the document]
Thog: It controls the way "out of bounds" time are handled (what rule need to be applied? the first one or the last one?)

time/src/main.rs
```diff
@@ -89,9 +89,10 @@ impl sunrise_libuser::time::StaticService for StaticService {
 #[derive(Debug)]
 struct Rtc {
     /// Command Register.
-    command: io::Pio<u8>,
+    command: Mutex<io::Pio<u8>>,
```
roblabla: Just for correctness' sake: This should be a single Mutex locking both `Pio`s. E.G. a `Mutex<(Pio, Pio)>`.

time/src/main.rs
```diff
@@ -244,21 +245,24 @@ fn main() {
...
+    RTC_INSTANCE.call_once(|| Rtc::new());
 
     let man = WaitableManager::new();
     let user_handler = Box::new(PortHandler::new("time:u\0", StaticService::dispatch).unwrap());
     let applet_handler = Box::new(PortHandler::new("time:a\0", StaticService::dispatch).unwrap());
     let system_handler = Box::new(PortHandler::new("time:s\0", StaticService::dispatch).unwrap());
     let rtc_handler = Box::new(PortHandler::new("rtc\0", RTCManager::dispatch).unwrap());
 
-    let rtc_instance = unsafe { Box::from_raw(&mut RTC_INSTANCE as *mut Rtc) };
+    //let rtc_instance = unsafe { Box::from_raw(&mut RTC_INSTANCE as *mut Rtc) };
```
roblabla: I think you can remove this :P

time/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+}
+
+fn main() {
+    // Setup a default device location
+    let device_location_name = b"Europe/Paris\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
+    timezone::TZ_MANAGER.lock().set_device_location_name(*device_location_name).unwrap();
+
+
+    RTC_INSTANCE.call_once(|| Rtc::new());
```
roblabla:
```suggestion
    let mut rtc = RTC_INSTANCE.call_once(|| Rtc::new());
```

time/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    let rtc_handler = Box::new(PortHandler::new("rtc\0", RTCManager::dispatch).unwrap());
+
+    //let rtc_instance = unsafe { Box::from_raw(&mut RTC_INSTANCE as *mut Rtc) };
+
+    man.add_waitable(user_handler as Box<dyn IWaitable>);
+    man.add_waitable(applet_handler as Box<dyn IWaitable>);
+    man.add_waitable(system_handler as Box<dyn IWaitable>);
+    man.add_waitable(rtc_handler as Box<dyn IWaitable>);
+
+    let mut rtc = RTC_INSTANCE.r#try().expect("RTC not initialized");
```
roblabla:
```suggestion
```

libtimezone/src/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    (((y) % 4) == 0 && (((y) % 100) != 0 || ((y) % 400) == 0))
+}
+
+/// Actual implementation of get_leap_days.
+#[inline]
+fn get_leap_days_not_neg(y: i64) -> i64 {
+    y / 4 - y / 100 + y / 400
+}
+
+/// Get the count of leap days of the given year.
```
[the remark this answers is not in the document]
Thog: Commenting here to not miss it: year 0 actually doesn't exist so that isn't an issue.

I might add a comment about it in the code just in case we miss that

time/src/timezone.rs
```diff
@@ -170,6 +170,10 @@ impl TimeZoneManager {
 /// This is a POD. There isn't any invariants so this should totally be safe.
 pub static TZ_MANAGER: Mutex<TimeZoneManager> = Mutex::new(unsafe { initialize_to_zero!(TimeZoneManager) });
 
+// Global clear instance of TimeZoneRule used to avoid copying 16KB on the stack.
+/// This is a POD. There isn't any invariants so this should totally be safe.
```
roblabla: I know this is a nit, but please, put this as a normal comment inside the unsafe block like this:

```rust
 static ZEROED_TIME_ZONE_RULE: TimeZoneRule = unsafe {
    // Safety: This is a POD. There isn't any invariants so this should totally be safe.
    initialize_to_zero!(TimeZoneRule)
};
```

I'm hoping to eventually build some tooling that will find unsafes that do not have such comments and resurface them.

time/src/timezone.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Get the device timezone rule.
+    pub fn get_my_rules(&self) -> &TimeZoneRule {
+        &self.my_rules
+    }
+}
+
+// https://data.iana.org/time-zones/tzdata-latest.tar.gz
+
+/// Global instance of TimeZoneManager
+pub static TZ_MANAGER: Mutex<TimeZoneManager> = Mutex::new(unsafe { initialize_to_zero!(TimeZoneManager) });
```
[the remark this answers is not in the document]
roblabla: Please put the safety note in the correct format, inside the `unsafe`  block.

libtimezone/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+            types: [0x0; TZ_MAX_TIMES as usize],
+            ttis: [TimeTypeInfo::default(); TZ_MAX_TYPES as usize],
+            chars: [0x0; 2 * (TZ_NAME_MAX as usize + 1)],
+            default_type: 0,
+            reserved: [0x0; 0x12c4],
+        }
+    }
+}
+
+// SAFE: TimeZoneRule satisfies all the requirements of `Plain`.
```
roblabla:
```suggestion
// Safety: TimeZoneRule satisfies all the requirements of `Plain`.
```

PR #322, opened 2019-06-13
Infra: creating the scripts folder

Orycterope:
This folder should only contain tools and utilities not essential to the compilation of SunriseOS, but which can help its debugging and development.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,3 @@
+### Content
+
+This folder should only contain tools and utilities not essential to the compilation of SunriseOS, but which can help its debugging and development.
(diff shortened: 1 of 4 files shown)


roblabla:
Slightly off-topic: The repository is starting to get a bit crowded. Maybe we should move "meta" non-code stuff to a dedicated folder, and keep only code at the top-level? e.g. have a `support` folder containing the `ipcdefs`, `scripts`, `external`, `linker-scripts`, and `scripts`? And, taking this a step further, move the crates into dedicated folders too? E.G. sysmodules/time, applets/wall-clock, applets/shell, etc...

Orycterope:
I'm all with it if we were to do both steps in one PR, but `support` is a really bad name.

Issue #314, opened 2019-06-07
paging - change initialisation


Orycterope:
When becoming the first process, we recycle the current page tables hierarchy as our process' page table hierarchy. 

When doing so, the kernel treats it as if it was empty, associating an empty bookkeeping with it. However, the bootstrap is still mapped in UserLand at this point.

This creates a situation where the bookkeeping does not reflect the state of the page tables, and leads to a kernel panic as soon as we try to map something in the UserLand of the init process (for me it was its TLS).

There are at least two straight-forward ways to solve this:

1. Make the bookkeeping reflect the state of the page tables. This means iterating the page tables and reserving any found frame in the `FrameAllocator`, the same way do for the KernelLand. 
2. Create a fresh ProcessMemory for the init process, and switch to it.

I have a preference for the second method, as it is simpler, even though it is a little bit less efficient.

PR #339, opened 2019-06-17
Thread local storage

Orycterope:
Adding support for Thread Local Storage (TLS) in userspace.

For every thread it creates, the kernel allocates a 0x200 bytes TLS region,
and makes `fs:0x0 .. fs:0x200` point to it.

For every thread it creates, the libuser allocates a TLS context,
stores its thread handle in it, and saves its address in its TLS region.

This PR defines the libuser low-level ABI to create threads,
and makes test_threads use it from now on.

~ELF Thread Local Storage will come as a separate PR, as this one
is required to implement mutexes, so we don't have to wait
to have full ELF TLS support to start development.~

Actually ELF TLS comes in this PR too, as it is closely linked to TLS region:

#### Userspace TLS:
For every thread it creates, the libuser copies the TLS initialization image, and makes `gs:0x0..gs:0xffffffff` point to it. The kernel provides a syscall for the libuser to set where the per-thread `gs` segment points to.
#### Kernel cpu-locals:
The kernel uses ELF TLS to implement cpu-local statics. For every core it detects, the kernel copies the kernel's TLS initialization image, and makes `gs:0x0..gs:0xffffffff` point to it. A cpu-local is therefore declared with the `#[thread_local]` attribute.

### `gs` switching
As the kernel and userspace both use `gs` to access thread_local/cpu_local statics, we switch where `gs` points to everytime we enter and leave the kernel. This is done via two segments in the gdt, one for userspace and one for kernelspace, and we switch which segment `gs` uses when we enter/leave the kernel.


Orycterope:
What's up with xmas_elf ? I'm panicking in the loading of vi since I merged master in this branch. Looks like xmas_elf is doing out of bounds accesses

Thog:
It's #314

kernel/src/i386/process_switch.rs
```diff
@@ -315,20 +320,26 @@ fn first_schedule() {
 /// This way, just after the `iret`, cpu will be in ring 3, witl all of its registers cleared,
 /// `$eip` pointing to `ep`, and `$esp` pointing to `userspace_stack_ptr`.
 fn jump_to_entrypoint(ep: usize, userspace_stack_ptr: usize, arg: usize) -> ! {
+    // gonna write constants in the code, cause not enough registers.
```
roblabla: So, *supposedly*, there's a way to use constants in asm blocks: `n` is "An integer constant – not including relocatable values.". I'd kinda be interested in seeing if those work (or if they make the compiler blow up).

kernel/src/paging/arch/i386/table.rs
```diff
@@ -299,6 +299,8 @@ impl InactiveHierarchyTrait for InactiveHierarchy {
         // Copy the kernel space tables
         self.copy_active_kernel_space();
         super::swap_cr3(self.directory_physical_address);
+        // Update the cr3 DOUBLE_FAULT_TSS will switch to when we double fault
+        crate::i386::gdt::DOUBLE_FAULT_TASK.lock().cr3 = self.directory_physical_address.addr() as u32;
```
roblabla: try_lock().expect() ?

Also, I'm not sure we should be touching DOUBLE_FAULT_TASK's cr3 at all. The double fault handler shouldn't be touching anything too fancy (like the heap or whatever). It should only access stuff that's already present on kernel startup and never unloaded.

kernel/src/panic.rs
```diff
@@ -0,0 +1,249 @@
...
+    unsafe { sync::permanently_disable_interrupts(); }
+    // Don't deadlock in the logger
+    unsafe { SerialLogger.force_unlock(); }
+
+    // Get the process we were running, and its name. Gonna be quite useful.
+    let current_thread = try_get_current_thread();
+    let current_process = current_thread.as_ref().map(|t| t.process.clone());
+    let current_process_name = current_process.as_ref().map(|p| &p.name);
+
+    //todo: force unlock the KernelMemory lock
```
roblabla: SMP makes this a non-starter.

kernel/src/i386/mod.rs
```diff
@@ -335,8 +342,17 @@ pub struct TssStruct {
...
+    /// Creates an empty TssStruct.
+    ///
+    /// All fields are set to `0`, suitable for static declarations, so that it can live in the `.bss`.
+    ///
+    /// The TssStruct must then be initialized with [init].
+    ///
+    /// Note that until it is initialized properly, the `.iopboffset` field will be invalid.
+    ///
+    /// [init]: TssStruct::init
+    pub const fn empty() -> TssStruct {
```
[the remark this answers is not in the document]
Orycterope: Actually lazy_static is a good idea, i'm going to try that.

However we can't put it behind a `#[thread_local]`, since it is used before cpu_locals are initialized, and I don't want to make the GDT's initialization too messy. 

Also, do we need one MAIN_TSS (and DOUBLE_FAULT_TSS) per cpu ? Since it's only used during double fault, i imagined one shared between cores would be enough, and we only need to retrieve the faulty core's hwcontext in the MAIN_TSS uppon double fault. This is based on the assumption that double faulting would stop SMP, so there's no risk of race on the TSS, but I haven't actually checked that assumption.

kernel/src/i386/process_switch.rs
Orycterope: The compiler is already blowing up with the `i` for no apparent reason on some generated asm blobs but not all, on i686-unknown-linux-gnu but not i386-unknown-none.

I'm going to try it, but I have no hopes at all.

kernel/src/paging/arch/i386/table.rs
Orycterope: I agree this is not ideal, but the cr3 we put initially in the the DOUBLE_FAULT_TASK is the directory allocated by the bootstrap, which will be freed when we become the first process (#315). This means that the double fault would likely page fault, since its cr3 points to a recycled page directory.

The problem comes from the fact that we have no canonical set of page tables that are updated every time we map/unmap in the kernel, we instead use the current page tables as the only source of truth, and copy this source of truth to another page directory when we process switch.

This current situation is not viable for SMP, and we would then have to properly implement this canonical set of page table as the new source of truth, shared and updated by every core. We could then make DOUBLE_FAULT_TASK's cr3 point to this directory.

In the mean time, the quick and dirty solution was to update DOUBLE_FAULT_TASK's cr3 everytime we process switch, which means the double fault handler can also access heap and start-up mapped structures without triple faulting. The overhead is just locking the DOUBLE_FAULT_TASK (spinlock) every time we process switch, which is not so terrible compared to all the overhead we have in the paging 👀.
The other solution would be to allocate a PageTableHierarchy just for the DOUBLE_TASK_HANDLER, copy early kernel space view to it, and never update it again. This lacks the advantages of the first solution, but I could go for this one if you really want to.

Orycterope: And lock() since we're never accessing the DOUBLE_FAULT_HANDLER anywhere else but in the double fault panic handler, so it can't possible be locked at that time. (unless we tried to process_switch in the double fault panic handler, which would be incredibly stupid idea) Do you want me to use try_lock().expect() instead and document it ?

kernel/src/panic.rs
Orycterope: Mmmmh 
tenor

kernel/src/i386/process_switch.rs
Orycterope: as expected: 
```
error: invalid operand for inline asm constraint 'n'
   --> kernel/src/interrupts/mod.rs:523:17
    |
523 | /                 asm!(trap_gate_asm!(has_errorcode: $errcode)
524 | |                 : /* no outputs */
525 | |                 : "s"($wrapper_rust_fnname as extern "C" fn (&mut UserspaceHardwareContext)),
526 | |                   "n"((GdtIndex::KTls as u16) << 3 | 0b000)
527 | |                 : "memory"
528 | |                 : "volatile", "intel");
    | |_______________________________________^
```

kernel/src/panic.rs
roblabla: KernelMemory could be locked by another CPU. Thinking about this more, we can force unlock kernel memory, but we need to shutdown all the other CPUs beforehand.

kernel/src/i386/process_switch.rs
roblabla: Hmm. There's a workaround, which is to put them in local constants, which would force the compiler to evaluate it before passing it to LLVM. E.G. instead of const asserts, doing:

```rust
const UCODE_IDX: u16 = (GdtIndex::UCode as u16) << 3 | 0b11;
// ...

asm!("" :: "n"(UCODE_IDX), ..);
```

But I guess `const_asserts` works too.

kernel/src/paging/arch/i386/table.rs
roblabla: >  The other solution would be to allocate a PageTableHierarchy just for the DOUBLE_TASK_HANDLER, copy early kernel space view to it, and never update it again. This lacks the advantages of the first solution, but I could go for this one if you really want to.

I might be missing something. What's the downside of this method compared to using the process' CR3? IMO, it'd be cleaner to go for this method. It's less state to keep track of and keep updating, which is always a good thing.

If, however, that's not desirable, then we should **definitely** go for `try_lock().expect()` (because if we deadlock there, then something went really wrong and is not recoverable - we want to kernel panic in this case).

Lastly, all those globals are actually cpu-local (as in, each CPU will need its own GDT, with its own double fault handler and TSS, etc...). Is using `#[thread_local]` for them possible at all? This would remove the need for a (totally unnecessary) mutex - replacing it with a RefCell.

If using `#[thread_local]` is not possible, I wonder if we could introduce an (unsafe) wrapper that allows us to treat them as CPU locals. Maybe a wrapper similar to `RefCell` which panics if accessed from the wrong CPU or something. Food for thought, not a blocker for this PR.

kernel/src/paging/arch/mod.rs
```diff
@@ -7,5 +7,5 @@ pub use self::i386::table::{ActiveHierarchy, InactiveHierarchy};
 pub use self::i386::entry::I386Entry as Entry;
 pub use self::i386::entry::I386EntryFlags as EntryFlags;
 pub use self::i386::is_paging_on;
-pub use self::i386::read_cr2; // todo give access to this in an arch-independent way
+pub use self::i386::{read_cr2, read_cr3}; // todo give access to this in an arch-independent way
```
Thog: Poor todobot
```suggestion
pub use self::i386::{read_cr2, read_cr3}; // TODO: give access to this in an arch-independent way
```

kernel/src/panic.rs
Orycterope: Yup, this is definitely what I had in mind. Panic should kill all cpus.

kernel/src/paging/arch/i386/table.rs
Orycterope: > I might be missing something. What's the downside of this method compared to using the process' CR3? IMO, it'd be cleaner to go for this method. It's less state to keep track of and keep updating, which is always a good thing.

We can't access heap or late-allocated static structures (like the GDT, which was in a lazystatic that allocated a page to store it, until now) in the double fault handler with this method. 
But you're right, it is less state to track and update. I'm considering doing this. 
But i want to insist that either way this is just a temporary design, until we have canonical page tables for SMP.

> If, however, that's not desirable, then we should definitely go for try_lock().expect() (because if we deadlock there, then something went really wrong and is not recoverable - we want to kernel panic in this case).

You're absolutely right on that.

> Lastly, all those globals are actually cpu-local (as in, each CPU will need its own GDT, with its own double fault handler and TSS, etc...). Is using #[thread_local] for them possible at all? This would remove the need for a (totally unnecessary) mutex - replacing it with a RefCell.

As I mentioned on another comment:

> However we can't put [the TSS] behind a #[thread_local], since it is used before cpu_locals are initialized, and I don't want to make the GDT's initialization too messy. This is the same reasoning behind the idea of the GDT itself being per-cpu, but can't be #[thread_local].

I agree that the mutex is unnecessary, and i'm trying to come up with a design that would work. However this is highly dependant on the way we start cores, and their startup sequence, which is not really clear for now. I see some solutions, but I'm afraid to implement them and later realize that they do not fit our startup sequence.

So for now i'm going to write a `// todo:` and leave it to a time when we'll be sure how the GDT/TSS allocation and init of a core happens. (parent ? child ? child runs on parent's GDT for a time, before creating its own ?)

kernel/src/i386/mod.rs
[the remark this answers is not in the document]
Orycterope: See https://github.com/sunriseos/SunriseOS/pull/339#discussion_r308305723

If we allocate one TSS per core, a simple lazy static doesn't work, as it needs to be called for every TSS we want to see initialized. If we manage to put those TSS behind a `#[thread_local]`, it seems to work inside lazy_static ([link to playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=55d55a025f5d6bbb8eeac26e865848fb)), but also seems like a bad idea. 

I'm waiting to know more clearly how other TSS are going to be allocated and initialized during the core startup sequence, in the mean time i'm adding a `// todo`.

Resolving conversation.

Orycterope:
Hhhm why didn't todo-bot pick up my todos ?

libkern/src/lib.rs
```diff
@@ -306,6 +310,50 @@ pub struct MemoryInfo {
...
+    /// to an address in the flat segmentation model that every other segment uses.
+    ///
+    /// This pointer serves as a translation.
+    pub ptr_self: *mut TLS,
+    /// reserved or unknown.
+    _reserved0: [u8; 16 - size_of::<*mut TLS>()],
+    /// Buffer used for IPC. Kernel reads, interprets, and copies data from/to it.
+    pub ipc_command_buffer: IpcBuffer,
+    /// reserved or unknown.
+    _reserved1: [u8; 0x200 - size_of::<IpcBuffer>() - 2 * size_of::<usize>() - (16 - size_of::<*mut TLS>())],
```
roblabla: This is sort of confusing. Why `2 * size_of::<usize>()` ? I suspect the answer is "the last pointer is at offset 0x198 on switch arm64", but well... who cares really :P.

libuser/src/crt0/relocation.rs
```diff
@@ -14,6 +14,8 @@ mod module_header {
         .int __bss_end__ - module_header
         .int __eh_frame_hdr_start__ - module_header
         .int __eh_frame_hdr_end__ - module_header
+        .int __tls_start__ - module_header
```
roblabla: This is not really RTLD-standard. I don't mind, but maybe add a comment that the RTLD doesn't put the TLS stuff here, since the rest is basically 1:1 nintendo stuff.

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// The alignment of the TLS segment.
+    ///
+    /// Because we don't want to read our own `P_TLS` program header,
+    /// the linker provides a symbol for the alignment it used.
+    ///
+    /// This is an **absolute symbol**, which means its "address" is actually its value,
+    /// i.e. to get its value do:
+    ///
+    /// ```
+    /// let tls_align = unsafe { &__tls_align__ as *const _ as usize };
```
roblabla: Ugh. https://github.com/rust-lang/rust/issues/36247. I hate this.

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+    // no dtv, no dynamics regions for now
+}
+
+impl TlsElf {
+    /// Allocates and initializes the static region, including TCB.
+    ///
+    /// Finds out the location of the initialization image from linker defined symbols.
+    pub fn allocate() -> Self {
+        // copy tls static area
+        let static_block_start = unsafe { (&module_header as *const _ as usize + module_header.tls_start as usize) as *const u8 };
```
roblabla:
```suggestion
        // TODO: Make module_header accesses safe
        // BODY: Accesses to the module_header static is totally safe - it is defined
        // BODY: by the linker and never mutated. However, all accesses to extern statics
        // BODY: are to be wrapped in unsafe blocks by default.
        // BODY: 
        // BODY: When https://github.com/rust-lang/rust/issues/36247 is fixed,
        // BODY: we should reinvestigate this to see if there's a way to tell rust that the access
        // BODY: is actually safe.
        let static_block_start = unsafe { (&module_header as *const _ as usize + module_header.tls_start as usize) as *const u8 };
```

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl ThreadLocalStaticRegion {
+    /// Returns a pointer to the [ThreadControlBlock] in the allocated region.
+    /// All TLS arithmetic are done relative to this pointer.
+    ///
+    /// For TLS to work, the value stored at this address should be the address itself, i.e.
+    /// having a pointer pointing to itself.
+    fn tcb(&self) -> *mut u8 {
+        unsafe {
+            // safe: guaranteed to be aligned, and still in the allocation
```
roblabla: avoid unsafe: `(self.ptr + self.tcb_offset) as *mut u8`

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+    fn allocate(block_src: &[u8], block_size: usize, block_align: usize) -> Self {
+        let tls_offset1 = tls_align_up(block_size, block_align);
+        let tcb_align = usize::max(align_of::<ThreadControlBlock>(), block_align);
+        let tcb_offset = tls_align_up(tls_offset1, tcb_align);
+        let alloc_pad_size = tcb_offset - tls_offset1;
+        let layout = Layout::from_size_align(
+            tcb_offset + size_of::<ThreadControlBlock>(),
+            tcb_align
+        ).unwrap();
+        let alloc = unsafe { alloc_zeroed(layout) };
```
roblabla: safety comment

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+        assert!(!alloc.is_null(), "thread_locals: failed static area allocation");
+
+        unsafe {
```
roblabla: safety comment

libuser/src/thread_local_storage.rs
```diff
@@ -0,0 +1,278 @@
...
+                tcb_offset
+            }
+        }
+    }
+}
+
+impl Drop for ThreadLocalStaticRegion {
+    /// Dropping a ThreadLocalStaticRegion deallocates it.
+    fn drop(&mut self) {
+        unsafe { dealloc(self.ptr as *mut u8, self.layout) };
```
roblabla: Safety comment

libuser/src/threads.rs
```diff
@@ -0,0 +1,290 @@
...
+///
+/// This function is the entry point of a thread, called directly by the kernel, with the
+/// argument passed by [Thread::create].
+/// It expects this argument to be the address of its `ThreadContext` so it can save it its `TLS`.
+///
+/// The routine to call and its argument are expected to be found in this `ThreadContext`.
+extern "fastcall" fn thread_trampoline(thread_context_addr: usize) -> ! {
+    debug!("starting from new thread, context at address {:#010x}", thread_context_addr);
+    // first save the address of our context in our TLS region
+    unsafe { (*get_my_tls_region()).ptr_thread_context = thread_context_addr };
```
roblabla: Safety comment.

Maybe `get_my_tls_region()` should return some kind of reference instead?

libuser/src/threads.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    let thread_context_addr = thread_context_addr as *mut ThreadContext;
+
+    // make gs point to our tls
+    let tls_elf = unsafe { &(*thread_context_addr).tls_elf };
```
roblabla: Safety comment.

libuser/src/threads.rs
```diff
@@ -0,0 +1,290 @@
...
+    tls_elf.r#try().unwrap().enable_for_current_thread();
+
+    // call the routine saved in the context, passing it the arg saved in the context
+    unsafe {
+        ((*thread_context_addr).entry_point)((*thread_context_addr).arg)
```
roblabla: Safety comment.

Actually, when you turn `thread_context_addr` into a raw pointer earlier in the function, maybe you should just turn it into a reference directly at that point. That'd contain the unsafety to that one point where it turns into a reference.

libuser/src/threads.rs
```diff
@@ -0,0 +1,290 @@
...
+///
+/// * put the main thread's handle in [MAIN_THREAD_CONTEXT].
+/// * save a pointer to it in its [TLS].
+/// * perform copy of `.tdata` and `.tbss` for the main thread.
+#[no_mangle] // called from asm
+pub extern fn init_main_thread(handle: ThreadHandle) {
+    // save the handle in our context
+    MAIN_THREAD_CONTEXT.thread_handle.call_once(|| handle);
+    // save the address of our context in our TLS region
+    unsafe { (*get_my_tls_region()).ptr_thread_context = &MAIN_THREAD_CONTEXT as *const ThreadContext as usize };
```
roblabla: safety comment.

kernel/src/i386/mod.rs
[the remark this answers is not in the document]
Orycterope: This is addressed as part of the issue in 7a030b6e04c9601235796ca75cd5644603d545f8

kernel/src/i386/process_switch.rs
```diff
@@ -255,44 +272,48 @@ pub unsafe fn prepare_for_first_schedule(t: &ThreadStruct, entrypoint: usize, us
...
+            main_tss.iopb[ioport / 8] &= !(1 << (ioport % 8));
         }
 
+        drop(main_tss); // unlock it
+
         // call the scheduler to finish the high-level process switch mechanics
-        let arg = current.arg;
-        crate::scheduler::scheduler_first_schedule(current, || jump_to_entrypoint(entrypoint, userspace_stack, arg));
+        unsafe {
+            // safety: interrupts are off
```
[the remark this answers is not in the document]
Orycterope: It's already in the doc of the function: https://github.com/sunriseos/SunriseOS/blob/af562899e7a7ee81991c01c07fb78808a04b5289/kernel/src/i386/process_switch.rs#L274

kernel/src/cpu_locals.rs
```diff
@@ -0,0 +1,344 @@
...
+impl CpuLocalRegion {
+    /// Returns a pointer to the [ThreadControlBlock] in the allocated region.
+    /// All cpu-local arithmetic are done relative to this pointer.
+    ///
+    /// For TLS to work, the value stored at this address should be the address itself, i.e.
+    /// having a pointer pointing to itself.
+    fn tcb(&self) -> *mut u8 {
+        unsafe {
+            // safe: guaranteed to be aligned, and still in the allocation
+            (self.ptr as *mut u8).add(self.tcb_offset)
```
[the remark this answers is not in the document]
Orycterope: It felt more intuitive, and the safety notice adds important information.

kernel/src/i386/gdt.rs
```diff
@@ -1,117 +1,294 @@
...
+//! The segment pointed by `gs` is controlled by the user. It can set its address/limit with
+//! [`svcSetThreadArea`]. The segment it chooses to use is local to every thread, and defaults to `0x00000000..0xffffffff`.
+//!
+//! Typically, the user will want to make `gs` point to its elf TLS.
+//!
+//! This segment is thread local, its address and size are switched at every thread-switch.
+//!
+//! ### LDT segments:
+//!
+//! None :)
```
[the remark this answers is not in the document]
Orycterope: Thought about it too, but the main idea of the LDT is to be per "task" (thread). Repurposing it for kernel vs. userspace isn't aligned with this. 

Moreover, I don't see any real improvement for doing this, and doing a `ldtr` would add the overhead of reloading every ldt-releated segment selectors (I think).

Add the fact that we have almost no support for the LDT, and the choice for GDT seems justified.

kernel/src/scheduler.rs
```diff
@@ -303,16 +324,33 @@ where
...
 
     unsafe {
         // this is a new process, no SpinLockIRQ is held
         crate::i386::instructions::interrupts::sti();
     }
 
+    // memset the TLS, to clear previous owner's data.
+    // we do it here so don't have to CrossProcessMap it earlier.
+    unsafe {
+        // safe: we manage this memory, ptr is aligned, and 0 is valid for every field of the TLS.
```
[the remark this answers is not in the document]
Orycterope: And exposing kernel padding bytes values isn't a security concern ? 🙃

libuser/src/threads.rs
Orycterope: No, because the IPC buffer needs to be borrowed mutably, and this can't be done trivially, we need to implement some kind of lock for it. So the only way to be safe for now is to return a *mut TLS. 

However I'm making `get_my_thread_context` return a const reference.

kernel/src/cpu_locals.rs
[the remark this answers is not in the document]
Orycterope: Actually, I should return a ref here.

libuser/src/threads.rs
roblabla: I mean, "we would need some kind of lock for it" holds true for a raw pointer as well - the compiler just doesn't enforce it. And we don't really need a lock, just some kind of Cell that ensures it's not borrowed multiple times.

libkern/src/lib.rs
Orycterope: Nop, just taking 0x200 bytes, and subtracting the size of every field in it one by one. The last pointer is at offset `0x200 - size_of::<usize>`.

I agree this was confusing, I'm simplifying it in the next commit.

libuser/src/crt0/relocation.rs
Orycterope: I'm removing it instead

libuser/src/thread_local_storage.rs
Orycterope: returning a safe &TCB instead

libuser/src/threads.rs
Orycterope: Yup. Even better: getting a const reference to the thread_context should always be safe, so I'm making `get_my_thread_context` do that.

kernel/src/cpu_locals.rs
```diff
@@ -0,0 +1,354 @@
...
+/// * Failed to get kernel ELF's TLS initialization image.
+pub fn init_cpu_locals(cpu_count: usize) {
+    debug_assert!(cpu_count > 0, "You can't have 0 cpu cores - I'm running code therefor I am");
+
+    CPU_LOCAL_REGIONS.call_once(|| {
+        // map our own ELF so that we can access our PT_TLS
+        let mapped_kernel_elf = multiboot::try_get_boot_information()
+            .and_then(|info| info.module_tags().nth(0))
+            .and_then(|module| map_grub_module(module).ok())
+            .expect("cpu_locals: cannot get kernel elf");
```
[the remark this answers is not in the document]
Orycterope: No it doesn't save any memory, as the initialisation image is in `.tdata`/`.tbss` whatever you do. 

This is the actual way you're **supposed** to do TLS. We use linker tricks in userspace because we don't want/can't access our own ELF, but in the kernel we can, so let's do it the normal way here.

libuser/src/crt0/relocation.rs
```diff
@@ -124,15 +128,15 @@ const R_386_RELATIVE: usize = 8;
 #[cfg(target_os = "none")]
 #[no_mangle]
 #[allow(clippy::cast_ptr_alignment)]
-pub unsafe extern fn relocate_self(aslr_base: *mut u8, module_header: *const ModuleHeader) -> u32 {
-    let module_header_address = module_header as *const u8;
-    let module_header = &(*module_header);
+pub unsafe extern fn relocate_self(aslr_base: *mut u8, module_headr: *const ModuleHeader) -> u32 {
```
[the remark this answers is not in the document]
Orycterope: they're supposed to be the same !

Renaming just so it doesn't conflict, and for clarity.

kernel/src/scheduler.rs
```diff
@@ -25,16 +28,18 @@ use crate::error::{UserspaceError};
...
 /// Will return None if we're in an early boot state, and it has not yet been initialized.
 pub fn try_get_current_thread() -> Option<Arc<ThreadStruct>> {
-    unsafe {
-        // Safe because modifications only happens in the schedule() function,
-        // and outside of that function, seen from a thread' perspective,
-        // CURRENT_THREAD will always have the same value
-        CURRENT_THREAD.clone()
+    // if cpu_locals haven't been initialized, accessing gs:0 will triple fault,
+    // so don't even remotely try to access it.
+    if !ARE_CPU_LOCALS_INITIALIZED_YET.load(Ordering::Relaxed) {
```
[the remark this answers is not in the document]
Orycterope: Other places would never be called in an early context.

I agree this is not ideal, I'm considering initializing TLS in bootstrap. This is not trivial however, and if I decide to it, i'll do it as a separate PR.

libuser/src/threads.rs
```diff
@@ -0,0 +1,320 @@
...
+///
+/// Internally owns the [ThreadContext] for this thread, including its stack.
+#[derive(Debug)]
+pub struct Thread(ManuallyDrop<Box<ThreadContext>>);
+
+impl Thread {
+    /// Start this thread.
+    pub fn start(&self) -> Result<(), Error> {
+        unsafe {
+            syscalls::start_thread(&(*self.0).thread_handle.r#try().unwrap())
```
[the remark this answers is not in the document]
Orycterope: This isn't even unsafe 🤦

PR #358, opened 2019-07-02
swipc-gen: Use enum_with_val! for enum

Thog:
This makes sure that enum are safe to be used with IPC.


swipc-gen/src/gen_rust_code.rs
```diff
@@ -390,8 +390,13 @@ fn format_cmd(cmd: &Func) -> Result<String, Error> {
 /// `TypeDef::Alias`, it will be a rust `type` alias.
 fn format_type(struct_name: &str, ty: &TypeDef) -> Result<String, Error> {
     let mut s = String::new();
-    for line in ty.doc.lines() {
-        writeln!(s, "/// {}", line).unwrap();
+
+    if let Type::Enum(_) = &ty.ty {
+        // Do nothing here
```
roblabla: Ugh, I wish there was a better way to do this...

Thog: ~~Where is the not keyword when we need it~~

PR #359, opened 2019-07-02
Make array buffer work in libuser 

Thog:
Also remove lifetime dependency


libuser/src/ipc/mod.rs
```diff
@@ -179,18 +179,53 @@ pub struct IPCBuffer<'a> {
 pub trait SizedIPCBuffer {
     /// Return the size of the type.
     fn size(&self) -> usize;
+
+    /// Check if the address and size are correct.
+    fn is_cool(addr: usize, size: usize) -> bool;
+
+    /// Create a reference to a ipc buffer from part.
```
roblabla:
```suggestion
    /// Create a reference to a ipc buffer from an address and a byte size.
```

Also might be worth linking to https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

libuser/src/ipc/mod.rs
```diff
@@ -179,18 +179,53 @@ pub struct IPCBuffer<'a> {
...
+    unsafe fn from_raw_parts<'a>(addr: usize, size: usize) -> &'a Self;
+
+    /// Create a mutable reference to a ipc buffer from part.
```
roblabla: Same

roblabla:
Thinking about this a bit more, we have no good reason to keep the buffer.rs file at all with the new object system. The `pop` functions should just return a reference (with a lifetime chosen by the caller).

Issue #332, opened 2019-06-15
Byteorder switch to `from_le_bytes`


Orycterope:
Since 1.32, u32 and other primitive types have a [from/to_le_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.from_le_bytes) method. 

Byteorder started to advocate it as a valid alternative to LE/BE in its doc. [see commit](https://github.com/BurntSushi/byteorder/commit/fd3fb623b3fbff689892a1641c99b4168076146b)

It seems that most (if not all) of our uses of byteorder could use the core functions instead, sparing us the dependency on byteorder.

roblabla:
We can't remove the byteorder dep entirely, because we have some functions (in libuser's PCI for instance) that are generic on the byte ordering. For the non-generic cases though, :+1:.

PR #360, opened 2019-07-02
Clean up usage of byteorder

Thog:
This remove dependency on byteorder in the kernel and shell.
Also reduce usage of byteorder in libuser

Related to #332


Orycterope:
We should fix #374 before merging

roblabla:
The cause for #374 was found, and is unrelated to this PR.

PR #363, opened 2019-07-08
Implement readable/writable events

roblabla:
Note: **This is 100% untested**. It compiles, but whether it works is totally up in the air. I'm pushing it here so I stop reimplementing it every fourth morning...


roblabla:
This is in #384

PR #377, opened 2019-07-16
Update nightly

roblabla:
Also add documentation about the updating process.


docs/README.md
```diff
@@ -0,0 +1,5 @@
+# SunriseOS Architecture Documentation
```
Orycterope: This "chapter" is not included in the generated docs, but really should.

However the way I would do it would be:

* Move this "SunriseOS Architecture Documentation" chapter to `/README.md`
* `/docs`'s generated docs includes `/README.md`. (this is already the case), including this chapter.
* Have `/docs/README.md` explain what the docs crate is for ("In the various submodules you'll find [..]") and how it works. This one is only displayed on github, and not included in the generated docs. A bit counter-intuitive, but the most straightforward design, which follows the`/scripts` convention.

docs/UPDATE_RUST.md
```diff
@@ -0,0 +1,22 @@
...
+    - clippy
+    - rls
+    - rust
+    - rust-analysis
+    - rust-docs
+    - rust-src
+    - rust-std
+    - rustc
+2. Set the `rust-toolchain` file at the root of the repo to `nightly-$LATEST_VER`. For instance, if the latest version to support all the requirements is 2019-07-15, then set the rust-toolchain file to `nightly-2019-07-15`.
+3. Compile the whole project, and fix all the errors.
```
Orycterope: Add a "Update cargo-xbuild" step.

docs/README.md
roblabla: I'm not sure what you mean? `/docs` doesn't generate from /README.md`, it generates from `docs/README.md`.

Orycterope: Hum ... no ? `/docs`'s generated documentation is `/README.md`:

Screen Shot 2019-07-17 at 1 33 28 PM

roblabla: Lol I pushed the wrong commit. Oh well.

roblabla: I think the main README.md should just point to the documentation like it currently does. It makes no sense for the docs to contain the main repo's readme, I pushed that by mistake when trying it out and being unsatisfied with the result.

Issue #374, opened 2019-07-12
xmas_elf OOB access in Kernel ELF loader


Thog:
This memory corruption only occurs while loading vi and can be reproduced on PR #360 .

roblabla:
https://github.com/sunriseos/SunriseOS/pull/360/commits/d9c80bfecae0080b0bd3f69b5a2763a34ceaec06 is the commit that's affected (in case PR gets force pushed)

Orycterope:
Can you give a description of the problem ?

Orycterope:
This bug is definitely present on master.
 
Haven't bisect yet, but reverting last xmas_elf update fixes the problem: 014dc189 (Jun 15 !)

I would like to investigate and report the bug to xmas_elf, but in the mean time I'm going to make a PR doing the revert.

Orycterope:
I was wrong, reverting xmas_elf update does no solve the problem. On some branches it seems to sometime disappear when doing a full rebuild of sunrise, hinting toward an incremental compilation problem.

For now I still believe this is a xmas_elf bug, and sunrise is not responsible for it. To prove my theory I'm going to try to replicate the OOB in a std crate that just tries to parse the "faulty" vi elf with xmas_elf.

roblabla:
@Thog found the problem: it's in mkiso-rs.

Once the bug is found, we should add a test to the CI that the generated ISO is correct by mounting the ISO and checking all the files.

Thog:
GRUB2 seems to not like the way my files were allocated, this should be fixed by the last release of mkisofs-rs. Please make sure to update it.

Orycterope:
Updated mkisofs-rs, diffed every module loaded by grub against their original elf, everything seems to be in order now. 

I'm closing the issue.

PR #385, opened 2019-07-21
Revert "Bump xmas-elf from 0.6.2 to 0.7.0"

Orycterope:
This reverts commit 014dc1891e4a35a2ec5743eeaf64a2bed270b780.

Addressing #374

xmas_elf does an out of bound access when loading vi, causing a kernel panic.

Reverting the last update of xmas_elf seems to fix the problem, use this until we figure out where this bug comes from.


Orycterope:
This commit doesn't solve the problem. xmas 0.6.2 > 0.7.0 are just trivial commits, and not the cause of the OOB. It appears the problem sometimes stops manifesting itself when doing a full rebuild, that's why this PR seemed to solve the problem, where actually it does not at all.

PR #390, opened 2019-07-29
Add a filesystem service

Thog:
This PR adds a filesystem service and changes AHCI to use IPC buffers.


disk-initializer/src/gpt.rs
```diff
@@ -0,0 +1,589 @@
...
+        self.partition_type = RawUUID::from_uuid(uuid);
+    }
+
+    /// Set the unique partition GUID.
+    pub fn set_unique_id(&mut self, uuid: Uuid) {
+        self.unique_id = RawUUID::from_uuid(uuid);
+    }
+
+    /// Set the name of the partition.
+    pub unsafe fn set_name(&mut self, name: &str) {
```
roblabla: Why is this unsafe? Under which condition is it safe to call this function? We need safety notes here.

disk-initializer/src/gpt.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    /// Read the content of a raw array into a GPTPartitionEntry.
+    pub fn read(&mut self, bytes: [u8; 0x80]) {
+        self.partition_type = RawUUID::from_slice_le(&bytes[0x0..0x10]);
+        self.unique_id = RawUUID::from_slice_le(&bytes[0x10..0x20]);
+        self.first_lba = LE::read_u64(&bytes[0x20..0x28]);
+        self.last_lba = LE::read_u64(&bytes[0x28..0x30]);
+        self.attribute = LE::read_u64(&bytes[0x30..0x38]);
+
+        let partition_name_u8 = unsafe {
```
roblabla: Safety note. You are casting a `RawUUID` to an array of `u8`. This is ok because RawUUID is repr(packed). If you turn `RawUUID` into `repr(C)`, it'll still be safe *so long as no padding bytes are present*.

Instead of this, I'd much rather we use `plain` again here.

disk-initializer/src/gpt.rs
```diff
@@ -0,0 +1,589 @@
...
+    pub fn write(&self) -> [u8; 0x80] {
+        let mut bytes = [0x0; 0x80];
+
+        (&mut bytes[0x0..0x10]).copy_from_slice(&self.partition_type.to_bytes_le());
+        (&mut bytes[0x10..0x20]).copy_from_slice(&self.unique_id.to_bytes_le());
+        LE::write_u64(&mut bytes[0x20..0x28], self.first_lba);
+        LE::write_u64(&mut bytes[0x28..0x30], self.last_lba);
+        LE::write_u64(&mut bytes[0x30..0x38], self.attribute);
+
+        let partition_name_u8 = unsafe {
```
roblabla: ditto. Use plain.

disk-initializer/src/gpt.rs
```diff
@@ -0,0 +1,589 @@
...
+        // Microsoft basic data GUID
+        main_partition
+            .set_partition_type(Uuid::parse_str("EBD0A0A2-B9E5-4433-87C0-68B6B72699C7").unwrap());
+
+        // Some GUID selected for the sake of randomness
+        main_partition
+            .set_unique_id(Uuid::parse_str("BA3E4ADC-EB06-11E7-8AD3-9570BEC474F8").unwrap());
+
+        // some name
+        unsafe { main_partition.set_name("SunriseOS System") };
```
roblabla: Safety note.

fs/src/detail/driver/fat/directory.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// The filter required by the user.
+    filter_fn: &'static dyn Fn(&FatFileSystemResult<FatDirectoryEntry>) -> bool,
+
+    /// The number of entries in the directory after ``filter_fn``.
+    entry_count: u64,
+}
+
+impl fmt::Debug for DirectoryInterface {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        unsafe {
```
roblabla: This is exactly why we shouldn't use `repr(packed)`. If possible, please use `repr(C)`. Otherwise, this implementation is in theory broken. I'm OK with punting the fix for later because we are sort of pressed for time and need to finish KFS-6 within short delays, but please add a TODO saying that we should rewrite this to avoid the UB.

fs/src/detail/driver/fat/file.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// The libfat's directory entry of this file.
+    file_inner: File,
+
+    /// File mode flags.
+    mode: FileModeFlags
+}
+
+impl fmt::Debug for FileInterface {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        unsafe {
```
roblabla: ditto. If possible repr(C), otherwise TODO.

fs/src/detail/driver/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+            if driver.is_supported(filesytem_type) {
+                return driver.format(storage, filesytem_type)
+            }
+        }
+
+        Err(FileSystemError::InvalidPartition.into())
+    }
+}
+
+unsafe impl Send for DriverManager {}
```
roblabla: Why is this necessary? Also, safety comment.

If this is a broken workaround, TODO.

fs/src/detail/driver/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        _ => panic!("{}", error)
+    }
+}
+
+impl BlockDevice for AhciDiskStorage {
+    /// Read blocks from the block device starting at the given ``index``.
+    fn read(&mut self, blocks: &mut [Block], index: BlockIndex) -> BlockResult<()> {
+        self.inner.read_dma(index.0, unsafe { core::slice::from_raw_parts_mut(blocks.as_mut_ptr() as *mut AhciBlock, blocks.len() ) }).map_err(|error| {
```
roblabla: Why is this safe?

If possible, we should have a function within the module that defines `AhciBlock` that transforms a `&mut [Block]` into a &mut [AhciBlock]`. See #391 about unsafety scope.

fs/src/detail/driver/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+            libuser_error_to_block_error(error, true)
+        })
+    }
+
+    /// Write blocks to the block device starting at the given ``index``.
+    fn write(&mut self, blocks: &[Block], index: BlockIndex) -> BlockResult<()> {
+        self.inner.write_dma(index.0, unsafe { core::slice::from_raw_parts(blocks.as_ptr() as *const AhciBlock, blocks.len() ) }).map_err(|error| {
```
roblabla: ditto

fs/src/detail/gpt/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+            last_lba: 0,
+            attribute: 0,
+            partition_name: [0x0; 0x24]
+        }
+    }
+}
+
+impl fmt::Debug for GPTPartitionEntry {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        unsafe {
```
roblabla: repr(C) or todo again.

fs/src/detail/gpt/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+        self.unique_id = RawUUID::from_uuid(uuid);
+    }
+
+    /// Get the unique partition GUID.
+    pub fn get_unique_id(&self) -> uuid::Uuid {
+        self.unique_id.to_uuid()
+    }
+
+    /// Set the name of the partition.
+    pub unsafe fn set_name(&mut self, name: &str) {
```
roblabla: Why is this unsafe? What are the invariants to uphold?

fs/src/detail/gpt/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// Read the content of a raw array into a GPTPartitionEntry.
+    pub fn read(&mut self, bytes: [u8; 0x80]) {
+        self.partition_type = RawUUID::from_slice_le(&bytes[0x0..0x10]);
+        self.unique_id = RawUUID::from_slice_le(&bytes[0x10..0x20]);
+        self.first_lba = LE::read_u64(&bytes[0x20..0x28]);
+        self.last_lba = LE::read_u64(&bytes[0x28..0x30]);
+        self.attribute = LE::read_u64(&bytes[0x30..0x38]);
+
+        let partition_name_u8 = unsafe {
+            core::slice::from_raw_parts_mut(self.partition_name.as_mut_ptr() as *mut u8, self.partition_name.len() * core::mem::size_of::<u16>())
```
roblabla: plain can do this safely.

fs/src/detail/gpt/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mut bytes = [0x0; 0x80];
+
+        (&mut bytes[0x0..0x10]).copy_from_slice(&self.partition_type.to_bytes_le());
+        (&mut bytes[0x10..0x20]).copy_from_slice(&self.unique_id.to_bytes_le());
+        LE::write_u64(&mut bytes[0x20..0x28], self.first_lba);
+        LE::write_u64(&mut bytes[0x28..0x30], self.last_lba);
+        LE::write_u64(&mut bytes[0x30..0x38], self.attribute);
+
+        let partition_name_u8 = unsafe {
+            core::slice::from_raw_parts(self.partition_name.as_ptr() as *const u8, self.partition_name.len() * core::mem::size_of::<u16>())
```
roblabla: plain again

fs/src/detail/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        let mut main_partition = GPTPartitionEntry::default();
+        // Microsoft basic data GUID
+        main_partition.set_partition_type(Uuid::parse_str("EBD0A0A2-B9E5-4433-87C0-68B6B72699C7").unwrap());
+
+        // Some GUID selected for the sake of randomness
+        main_partition.set_unique_id(Uuid::parse_str("BA3E4ADC-EB06-11E7-8AD3-9570BEC474F8").unwrap());
+
+        // some name
+        unsafe { main_partition.set_name("SunriseOS System") };
```
roblabla: unsafe again

fs/src/detail/gpt/mod.rs
Thog: urgh it used to do unsafe stuffs before but not anymore oops

Thog:
@roblabla Removing the code duplication entirely isn't possible at all as most of the type needed by this code depends on libuser that cannot be used in the host std context.

I can still move th gpt part but that's it.

fs/src/detail/driver/fat/directory.rs
Thog: no repr here, why have I made this unsafe 🤔

fs/src/detail/driver/fat/file.rs
Thog: same as previous comment 🤔

disk-initializer/src/gpt.rs
```diff
@@ -202,10 +202,8 @@ impl GPTPartitionEntry {
         self.attribute = LE::read_u64(&bytes[0x30..0x38]);
 
         let partition_name_u8 = unsafe {
-            core::slice::from_raw_parts_mut(
-                self.partition_name.as_mut_ptr() as *mut u8,
-                self.partition_name.len() * core::mem::size_of::<u16>(),
-            )
+            // Safety: array of u16 can be represented as array of u8 so this is safe.
```
roblabla: Safety comment can go away, there's no unsafe code here :').

disk-initializer/src/gpt.rs
```diff
@@ -222,10 +220,8 @@ impl GPTPartitionEntry {
         LE::write_u64(&mut bytes[0x30..0x38], self.attribute);
 
         let partition_name_u8 = unsafe {
-            core::slice::from_raw_parts(
-                self.partition_name.as_ptr() as *const u8,
-                self.partition_name.len() * core::mem::size_of::<u16>(),
-            )
+            // Safety: array of u16 can be represented as array of u8 so this is safe.
```
roblabla: ditto

kernel/src/main.rs
```diff
@@ -147,9 +147,9 @@ fn main() {
                 let stack = pmemlock.find_available_space(5 * PAGE_SIZE)
```
roblabla:
```suggestion
                let stack = pmemlock.find_available_space(9 * PAGE_SIZE)
```

ipcdefs/ahci.id
```diff
@@ -1,3 +1,6 @@
+# A block (or sector) used by AHCI I/O operations.
+type sunrise_libuser::ahci::Block = bytes<512>;
```
Orycterope: This must be aligned to 2, see AHCI PRDT requirements.

ahci/src/disk.rs
```diff
@@ -186,12 +186,8 @@ impl IDiskInterface for IDisk {
...
     ///       the PRDT. This can only happen for read/writes of 1985 sectors or more.
     ///       You should consider retrying with a smaller `sector_count`.
-    fn read_dma(&mut self, _manager: &WaitableManager, sharedmem: SharedMemory, mapping_size: u64, lba: u64, sector_count: u64) -> Result<(), Error> {
-        let addr = sunrise_libuser::mem::find_free_address(mapping_size as _, 0x1000)?;
-        let mapped = sharedmem.map(addr, mapping_size as _, MemoryPermissions::empty())
-        // no need for permission, only the disk will dma to it.
-            .map_err(|_| AhciError::InvalidArg)?;
-        self.0.lock().read_dma(mapped.as_mut_ptr(), mapped.len(), lba, sector_count)
+    fn read_dma(&mut self, _manager: &sunrise_libuser::ipc::server::WaitableManager, address: u64, out_blocks: &mut [sunrise_libuser::ahci::Block]) -> Result<(), Error> {
+        self.0.lock().read_dma(out_blocks.as_mut_ptr() as *mut u8, out_blocks.len() * core::mem::size_of::<Block>(), address, out_blocks.len() as u64)
```
Orycterope: We need to check the alignment here, when creating a pointer.

ahci/src/disk.rs
```diff
@@ -104,7 +104,7 @@ impl Disk {
     ///
     /// Writes `sector_count` sectors starting from `lba`.
     #[inline(never)]
-    fn write_dma(&mut self, buffer: *mut u8, buffer_len: usize, lba: u64, sector_count: u64) -> Result<(), Error> {
+    fn write_dma(&mut self, buffer: *const u8, buffer_len: usize, lba: u64, sector_count: u64) -> Result<(), Error> {
```
Orycterope: You need to update the unsafe invariants in `read_dma` and `write_dma`.

Just update the comments.

ahci/src/disk.rs
Orycterope: It might even be a good idea to create a &mut [Block] here, ensuring all is good, size and alignment, and have the lower-level functions become safe.

ipcdefs/filesystem.id
```diff
@@ -0,0 +1,168 @@
...
+    # Rename a file at ``old_path`` into ``new_path``.
+    [5] rename_file(buffer<sunrise_libuser::fs::FileSystemPath, 0x19, 0x301> old_path, buffer<sunrise_libuser::fs::FileSystemPath, 0x19, 0x301> new_path);
+
+    # Rename a directory at ``old_path`` into ``new_path``
+    [6] rename_directory(buffer<sunrise_libuser::fs::FileSystemPath, 0x19, 0x301> old_path, buffer<sunrise_libuser::fs::FileSystemPath, 0x19, 0x301> new_path);
+
+    # Get the informations about an entry on the filesystem.
+    [7] get_entry_type(buffer<sunrise_libuser::fs::FileSystemPath, 0x19, 0x301> path) -> sunrise_libuser::fs::DirectoryEntryType;
+
+    # Open a file at the specified ``path`` with the given ``mode`` flags.
```
roblabla:
```suggestion
    # Open a file at the specified ``path`` with the given ``mode`` flags.
    #
    # ``mode`` controls how the file is opened, based on which bits are set:
    # When bit 0 is set, the file is Readable: you can use the Read operation.
    # When bit 1 is set, the file is Writable: you can use the Write operation.
    # When bit 2 is set, the file is Appendable: unless this bit is set, you will not be able to write beyond the end of a file (such writes will result in an error 0x307202)
```

ipcdefs/filesystem.id
```diff
@@ -0,0 +1,168 @@
...
+    # Read the next directory entries and return the number of entries read.
+    [0] read() -> (u64, array<sunrise_libuser::fs::DirectoryEntry, 0x6>);
+
+    # Return the count of entries in the directory.
+    [1] get_entry_count() -> u64;
+}
+
+# Represent a file in a filesystem.
+interface sunrise_libuser::fs::IFile {
+    # Read the content of a file at a given ``offset`` in ``out_buf``.
```
roblabla:
```suggestion
    # Read the content of a file at a given ``offset`` in ``out_buf``. ``option`` should be set to 0.
```

ipcdefs/filesystem.id
```diff
@@ -0,0 +1,168 @@
...
+    [0] read(u32, u64 offset, u64 size) -> (u64 out_size, array<u8, 0x46> out_buf);
```
roblabla:
```suggestion
    [0] read(u32 option, u64 offset, u64 size) -> (u64 out_size, array<u8, 0x46> out_buf);
```

ipcdefs/filesystem.id
```diff
@@ -0,0 +1,168 @@
...
+
+    # Write the content given ``in_buf`` at the given ``offset`` in the file.
+    # If the file is too small to hold the data and the appendable flag is set, it will resize the file and append the data.
+    # If the file is too small to hold the data and the appendable flag isn't set, this will return a FileSystemError::NoSpaceLeft.
```
roblabla:
```suggestion
    # If the file is too small to hold the data and the appendable flag isn't set, this will return a FileSystemError::NoSpaceLeft.
    # ``option`` should be set to 0.
```

ipcdefs/filesystem.id
```diff
@@ -0,0 +1,168 @@
...
+    [1] write(u32, u64 offset, u64 size, array<u8, 0x45> in_buf);
```
roblabla:
```suggestion
    [1] write(u32 option, u64 offset, u64 size, array<u8, 0x45> in_buf);
```

external/filesystem/disk_template/etc/motd
```diff
@@ -0,0 +1,3 @@
+Welcomme to Sunrise OS
```
Orycterope:
```suggestion
Welcome to Sunrise OS
```

images-1

shell/src/main.rs
```diff
@@ -40,9 +42,42 @@ use alloc::vec::Vec;
 use alloc::boxed::Box;
 use alloc::sync::Arc;
 use spin::Mutex;
+use bitflags::bitflags;
+
+bitflags! {
+    /// Flags indicating the way a file should be open.
+    pub struct FileModeFlags: u32 {
```
Orycterope: This should live in libuser, in a /src/fs.rs module perhaps

fs/src/interface/filesystem.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn read(&mut self, buf: &mut [DirectoryEntry]) -> LibUserResult<u64>;
+
+    /// Return the count of entries in the directory.
+    fn entry_count(&self) -> LibUserResult<u64>;
+}
+
+/// Represent the operation on a filesystem.
+pub trait FileSystemOperations : core::fmt::Debug + Sync + Send {
+    /// Create a file with a given ``size`` at the specified ``path``.
+    fn create_file(&self, path: &str, size: u64) -> LibUserResult<()>;
```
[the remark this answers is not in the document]
Orycterope: Edit: 

An open File handle is an outstanding reference to the filesystem, but you can't tie the lifetimes. 

This means the driver must have a mutex around its filesystem, and opt for the second strategy.

fs/src/interface/storage.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        // copy the needed content
+        buf.copy_from_slice(&self.temp_storage[0][current_block_offset .. current_block_offset + buf.len()]);
+
+        Ok(())
+    }
+}
+
+impl<B> StorageDevice for StorageCachedBlockDevice<B> where B: BlockDevice + Sync + Send {
+    fn read(&mut self, offset: u64, buf: &mut [u8]) -> StorageDeviceResult<()> {
```
Orycterope: Ok, multiple things are wrong with this function.

# Misaligned read

First, you cannot possibly write to an unaligned buffer without a shifting memcpy for the middle part. 

### The simple case 

To demonstrate that, let's consider only the middle part to make things simple, i.e. we're simply reading a multiple of Block size to an unaligned buffer. What we're trying to accomplish is something like this:

```
                 block 0 bytes      block 1 block 2
disk:       [ 0 | 1 | 2 | 3 | ...][  ...  ][  ...  ]
            /
           V      
buffer: [ 0 | 1 | 2 | 3 | ... ]
```

You can see that the buffer and the disk are horizontally misaligned, to represent their misalignment in memory. 

What you're currently doing is is reading the first byte on its own, and then fill the rest of the buffer directly from the disk, effectively doing this:

```
disk:       [ 0 | 1 | 2 | 3 | ...][  ...  ][  ...  ]
              |
              V   
buffer: [ 0 | 0 | 1 | 2 | 3 | ... ]
```

You can see that the first byte of the block is present twice. This is wrong, but unfortunately I see no way to simply read the block and expect it to work.

We always have to first read from the disk in an aligned fashion, and then shift everything in place to misalign it to our buffer misalignment:

```
disk:       [ 0 | 1 | 2 | 3 | ...][  ...  ][  ...  ]
              |
              V   
buffer: [   | 0 | 1 | 2 | 3 | ... ]
memcpy:   V---┘
buffer: [ 0 | 1 | 2 | 3 | ... ... ]
```

But for us to read and then shift in place, our buffer must be big enough to hold all the blocks, plus some padding for the shift. In practice this means that often a read of 3 blocks would only read 2 into the buffer so it has enough room, do the shift, read the 3rd block into a tmp aligned block, and copy from that aligned tmp to the end of the misaligned buffer. 

### The real life case

Until now we've only considered the simple case of reading multiples of Block size. But most of the time we're not. To do that we still split the problem into a "before, middle, after" three time read. For the before and after parts, reading to an aligned tmp buffer and memcpy to misaligned buffer is not a problem. However the middle still needs to be shifted. **But** we can potentially harvest the "after" part in the buffer as extra free space for our shift, and save one read. This can only happer if there's an "after" part however, if the buffer ends on a multiple of block size we can't :/ 

# Nothing to do with fs

Finally, this is a lot more related to storage_device than it is to fs, so it should live in storage_device. I was thinking of making storage_device a public crate on crates.io. Blocks of size 512 and aligned to 2 is very specific to ahci, but we could define a whole family of different blocks and let the user define its own, with its own alignement, and be generic over that.

# Time constraint 

Even though the current implementation is broken af, because we're short on time, if you can guarantee me that it seems to be working for now (maybe we're lucky and we're never reading to unaligned buffers, thanks rustc), I will let it pass and be merged for now. 

One of the first things I'll do will be to work on a rewrite of all of this in storage_device, and make storage_device publishable.

Also I'll write lots of unit tests for my storage_device impl, as we seem to have a hard time getting it right.

disk-initializer/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+use libfat::FatFsType;
+use libfat::directory::File as FatFile;
+
+mod gpt;
+
+use gpt::{PartitionIterator, PartitionManager};
+
+/// A structure used to manipulate std::File as BlockDevice
+#[derive(Debug)]
+struct StdBlockDevice {
```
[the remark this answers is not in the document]
Thog: totally missed that lol

disk-initializer/src/gpt.rs
```diff
@@ -0,0 +1,584 @@
...
+        let mut partition_table = Vec::new();
+
+        let mut main_partition = GPTPartitionEntry::default();
+        // Microsoft basic data GUID
+        main_partition
+            .set_partition_type(Uuid::parse_str("EBD0A0A2-B9E5-4433-87C0-68B6B72699C7").unwrap());
+
+        // Some GUID selected for the sake of randomness
+        main_partition
+            .set_unique_id(Uuid::parse_str("BA3E4ADC-EB06-11E7-8AD3-9570BEC474F8").unwrap());
```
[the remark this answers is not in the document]
Thog: UUIDs aren't the same :P

fs/src/detail/driver/fat/directory.rs
```diff
@@ -0,0 +1,175 @@
...
+    base_path: [u8; PATH_LEN],
+
+    /// libfat filesystem interface.
+    inner_fs: Arc<Mutex<libfat::filesystem::FatFileSystem<PartitionStorage>>>,
+
+    /// The iterator used to iter over libfat's directory entries.
+    internal_iter: FatDirectoryEntryIterator,
+
+    /// The filter required by the user.
+    filter_fn: fn(&FatFileSystemResult<FatDirectoryEntry>) -> bool,
```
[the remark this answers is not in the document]
Thog: This complexity the code quite a lot especially on the entry_count computation so I will keep this that way.

fs/src/interface/filesystem.rs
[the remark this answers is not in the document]
Thog: Second strategy is already in place 👍

fs/src/detail/driver/fat/directory.rs
```diff
@@ -0,0 +1,175 @@
...
+        for (index, entry) in buf.iter_mut().enumerate() {
+            let mut raw_dir_entry;
+            loop {
+                let filesystem = self.inner_fs.lock();
+                let entry_opt = self.internal_iter.next(&filesystem);
+
+                // Prematury ending
+                if entry_opt.is_none() {
+                    return Ok(index as u64);
+                }
```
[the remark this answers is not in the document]
Thog: This cannot be done as ``to_iterator`` requires to consume ``internal_iter`` :/

fs/src/detail/driver/fat/directory.rs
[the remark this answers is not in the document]
Orycterope: Hum, no. If you get an error reading, return it.

Otherwise you're returning Ok, empty folder. Which it is not ! Even worse, if the folder is empty, this means we can rmdir it now, great !

And now the user sees an empty folder, but can't delete it. Panic incoming.

shell/src/main.rs
```diff
@@ -32,24 +32,75 @@ extern crate sunrise_libuser as libuser;
...
+                            let _ = writeln!(&mut terminal, "cat: {}", error);
+                        }
+                    }
+                }
+
+            }
+            "pwd" => {
+                let _ = writeln!(&mut terminal, "{}", CURRENT_WORK_DIRECTORY.lock().as_str());
+            },
+            "cd" => {
```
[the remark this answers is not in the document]
roblabla: :+1: to dir.

PR #415, opened 2019-08-12
todo-bot: touch TLS todos

Orycterope:
Because the TLS PR is so big, todobot bailed on it.

Make a PR that only touches every `// todo:` we opened in #339, so todobot picks them again.


roblabla:
This case swap is a terrible hack. I love it.

PR #433, opened 2019-08-14
Update to cargo-make 0.22.0

### LINKED ISSUE #430: Update to cargo-make 0.22.0
Cargo-make 0.22.0 adds new interesting shit, notably:

- Auto-update dependencies that don't meet a minimum version
- Globally disable workspace support via makefile config (gone are all the ugly `workspace = false`)
- Support additional rust script runners (so we can write rust tasks without breaking windows builds in the future)

roblabla:
It will now update our deps. Fixes #430


Makefile.toml
```diff
@@ -216,22 +199,20 @@ args = [ "file", "createnew", "DISK.img", "8388608" ]
 
 # because we can't have a dependency on a file ರ_ರ
 [tasks.create-disk-if-not-exist]
-workspace = false
 description = "Invokes the task `disk` if DISK.img does not already exist."
-condition_script = [ "[ ! -e DISK.img ]" ]
+#condition_script = [ "[ ! -e DISK.img ]" ]
```
Orycterope: Just find a way to re-enable this, and it'll be good.

roblabla: Lol, that change is part of `loader`, it snuck into this PR. Will fix in a moment.

PR #448, opened 2019-08-25
Shell password

roblabla:
Adds a password to the shell. This checks the `/etc/passwd` file, which contains the username and an SHA-1 hashed password (not salted yet. I sort of lacked time and this feature is only there to satisfy the KFS project anyways).

Also adds a useradd command which is useful to showcase writing to the filesystem.


external/filesystem/disk_template/etc/passwd
```diff
@@ -0,0 +1,2 @@
+roblabla 7288edd0fc3ffcbe93a0cf06e3568e28521687bc
+orycterope 35675e68f4b5af7b995d9205ad0fc43842f16450
```
Thog:
```suggestion
orycterope 35675e68f4b5af7b995d9205ad0fc43842f16450
thog 1af9edbb9f29c29f469cf4ff9c15b7e8397ac33f
```

PR #450, opened 2019-08-25
Implement ls

roblabla:
EzPz.

### DIFF
--- a/shell/src/main.rs
+++ b/shell/src/main.rs
@@ -260,10 +260,17 @@ fn cat<W: Write>(f: &mut W, filesystem: &IFileSystemProxy, file: &str) -> Result
 
 /// List files and folders at the given path, or in the current path is none is
 /// given.
-fn ls(mut terminal: &mut Terminal, filesystem: &IFileSystemProxy, path: Option<&str>) -> Result<(), Error> {
+fn ls(mut terminal: &mut Terminal, filesystem: &IFileSystemProxy, orig_path: Option<&str>) -> Result<(), Error> {
     use sunrise_libuser::fs::{DirectoryEntry, DirectoryEntryType};
 
-    let path = path.unwrap_or(".");
+    #[allow(clippy::missing_docs_in_private_items)]
+    const BG: Color = Color::rgb(0, 0, 0);
+    #[allow(clippy::missing_docs_in_private_items)]
+    const FILE_FG: Color = Color::rgb(0x56, 0xF7, 0x68);
+    #[allow(clippy::missing_docs_in_private_items)]
+    const DIR_FG: Color = Color::rgb(0x68, 0x70, 0xF1);
+
+    let path = orig_path.unwrap_or(".");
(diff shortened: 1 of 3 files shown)


shell/src/main.rs
```diff
@@ -254,6 +258,47 @@ fn cat<W: Write>(f: &mut W, filesystem: &IFileSystemProxy, file: &str) -> Result
...
+    let path = path.unwrap_or(".");
+    let path = get_path_relative_to_current_directory(path);
+    if path.len() > 0x300 {
+        return Err(FileSystemError::InvalidInput.into())
+    }
+
+    let mut ipc_path = [0x0; 0x300];
+    ipc_path[..path.as_bytes().len()].copy_from_slice(path.as_bytes());
+
+    let directory = filesystem.open_directory(3, &ipc_path)?;
```
Orycterope: what about 
```
ls some_file
```
 ?

shell/src/main.rs
```diff
@@ -254,6 +258,47 @@ fn cat<W: Write>(f: &mut W, filesystem: &IFileSystemProxy, file: &str) -> Result
     Ok(())
 }
 
+/// List files and folders at the given path, or in the current path is none is
+/// given.
+fn ls(mut terminal: &mut Terminal, filesystem: &IFileSystemProxy, path: Option<&str>) -> Result<(), Error> {
+    use sunrise_libuser::fs::{DirectoryEntry, DirectoryEntryType};
+
+    let path = path.unwrap_or(".");
+    let path = get_path_relative_to_current_directory(path);
```
Orycterope: does this work for absolute paths ?

shell/src/main.rs
```diff
@@ -254,6 +258,47 @@ fn cat<W: Write>(f: &mut W, filesystem: &IFileSystemProxy, file: &str) -> Result
...
+        for entry in entries {
+            let split_at = entry.path.iter().position(|v| *v == 0).unwrap_or(0x300);
+            let prefix_len = if path == "/" {
+                // The prefix to remove is literally just the leading `/`
+                1
+            } else {
+                // The prefix to remove is the path + the trailing `/`.
+                path.len() + 1
+            };
+            let _ = writeln!(&mut terminal, "{}", String::from_utf8_lossy(&entry.path[prefix_len..split_at]));
```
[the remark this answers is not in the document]
Orycterope: Or color in green, the logger can do it IIRC

shell/src/main.rs
roblabla: I checked, and the answer is yes.

shell/src/main.rs
roblabla: Currently returns an error saying it's not a directory. Will fix tomorrow.

shell/src/main.rs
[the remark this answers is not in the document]
roblabla: Colors is probably a better idea. Will take a look tomorrow, I think Terminal can output some colors.

PR #460, opened 2019-08-27
Switch to multiboot2 official releases

Thog:
We don't really have a reason to use a custom fork.


bootstrap/src/frame_alloc.rs
```diff
@@ -160,12 +160,9 @@ impl FrameAllocator {
             if memarea.start_address() > u32::max_value() as u64 || memarea.end_address() > u32::max_value() as u64 {
                 continue;
             }
-            if memarea.memory_type() == 1 {
```
roblabla: :eyes: What did that used to do, and why is this not necessary anymore?

Thog: I removed this because we had a custom version of this iterator in our fork that return all memory area with their memory type instead of only the FREE one

PR #465, opened 2019-08-28
Mutex

Orycterope:


kernel/src/i386/gdt.rs
```diff
@@ -493,7 +493,7 @@ impl MainTask {
 // BODY: to the current cr3, and update it when we switch page table hierarchies. However the current
 // BODY: way we do kernel paging is not viable for SMP, and we might finally implement such a hierarchy
 // BODY: for SMP, we could then make DOUBLE_FAULT TSS(s) point to it.
-pub static MAIN_TASK: Mutex<MainTask> = Mutex::new(MainTask::empty());
+pub static MAIN_TASK: SpinLock<MainTask> = SpinLock::new(MainTask::empty());
```
roblabla: The sad truth: This should really be thread_local :')

kernel/src/main.rs
```diff
@@ -10,7 +10,7 @@
 //! Currently doesn't do much, besides booting and printing Hello World on the
 //! screen. But hey, that's a start.
 
-#![feature(lang_items, start, asm, global_asm, compiler_builtins_lib, naked_functions, core_intrinsics, const_fn, abi_x86_interrupt, allocator_api, box_syntax, no_more_cas, const_vec_new, step_trait, thread_local, nll, doc_cfg, exclusive_range_pattern)]
+#![feature(lang_items, start, asm, global_asm, compiler_builtins_lib, naked_functions, core_intrinsics, const_fn, abi_x86_interrupt, allocator_api, box_syntax, no_more_cas, const_vec_new, step_trait, thread_local, nll, doc_cfg, exclusive_range_pattern, optin_builtin_traits)]
```
roblabla: I suppose you use this in order to add !Send !Sync to your Mutex? The safe way to do it is to add a `PhantomData<*mut ()>`, which is !Send !Sync.

kernel/src/sync/mutex.rs
```diff
@@ -0,0 +1,667 @@
...
+/// [`DerefMut`]: core::ops::DerefMut
+/// [`lock`]: Mutex::lock
+/// [`try_lock`]: Mutex::try_lock
+#[must_use = "if unused the Mutex will immediately unlock"]
+pub struct MutexGuard<'a, T: 'a> {
+    /// Reference to the Mutex we'll unlock when dropped.
+    __lock: &'a Mutex<T>,
+}
+
+impl<T> !Send for MutexGuard<'_, T> { }
```
roblabla: Put a `PhantomData<*mut ()>` in the struct instead.

kernel/src/i386/gdt.rs
Orycterope: Yes 😭

kernel/src/main.rs
Orycterope: Oh ... ok ... really hacky

PR #472, opened 2019-09-13
Rewrite timer to support oneshot

Thog:
This PR improve the timer implementation that we currently have by switching to a oneshot approach.

Also remove PIT support and improve HPET driver safety.

### DIFF
--- a/kernel/src/timer.rs
+++ b/kernel/src/timer.rs
@@ -127,7 +127,7 @@ impl Waitable for TimerEvent {
         let mut target_ticks = self.target_ticks.lock();
         let result = {
             let timer_driver = TIMER_DRIVER.r#try().expect("Timer driver is not initialized!").lock();
-            timer_driver.is_after_or_equal_main_ticks(*target_ticks)
+            !timer_driver.is_after_or_equal_main_ticks(*target_ticks)
         };
 
         if result {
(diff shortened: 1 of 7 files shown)


kernel/src/timer.rs
```diff
@@ -1,103 +1,191 @@
...
+    for _ in target_index {
+        // As the vec is always sorted, we just remove the first element every time
+        let state = states.remove(0);
+        crate::scheduler::add_to_schedule_queue(state.target_waiter);
+    }
+
+    let next_oneshot = states.get(0);
+
+    // Setup next oneshot if present.
+    if let Some(next_oneshot) = next_oneshot {
```
roblabla: I'm pretty sure we can lose interrupts here. What happens if between the `is_after_or_equal` and this point, target_ticks is passed?

We should be doing the whole `wakeup_waiters` logic in a loop. After setting up the oneshot timer, check if `timer_driver.is_after_or_equal_target_ticks`, and if it isn't, rerun from `let target_index: Vec<usize>`.

kernel/src/timer.rs
```diff
@@ -165,27 +166,39 @@ fn register_timer_event(event: &TimerEvent) {
...
-    // TODO: mask interruptions?
     let mut timer_driver = TIMER_DRIVER.r#try().expect("Timer driver is not initialized!").lock();
 
     let mut states = TIMER_STATES.lock();
 
-    // Get all threads to wake up.
-    let target_index: Vec<usize> = states.iter().enumerate().filter(|x| timer_driver.is_after_or_equal_target_ticks(x.1.target_ticks)).map(|x| x.0).collect();
+    loop {
+        // Get all threads to wake up.
+        let next_oneshot_idx = states.iter().position(|x| !timer_driver.is_after_or_equal_main_ticks(x.target_ticks)).unwrap_or_else(|| states.len());
```
roblabla: This is wrong. We want the position of the first element whose target tick is after the main tick.

Issue #468, opened 2019-09-09
Stop using GRUB


roblabla:
GRUB is written in the unholy C language, and as such should be expunged from our systems! In order to implement secureboot, we need to run `grub-mkimage` in order to embed a key in the GRUB bootloader and embed our grub config. This is annoying as it breaks the build for windows users.

We should use our own bootloader. This bootloader should closely follow the design of NX-Bootloader (the bootloader of HOS/NX). Which means it should:

1. Do some amount of hardware initialization
2. Find and load the Package2 (containing the Kernel + builtins)
3. Verify its signature against a hardcoded key (for simplicity's sake, will be the same as the secure boot key)
4. Jump to the kernel's entrypoint.

This bootloader should compile to an UEFI binary (rust natively supports building UEFI binaries) that will get signed by the secure boot key.

roblabla:
The most complicated part of this setup will be setting things up to get ourselves a framebuffer. Maybe we should move the framebuffer acquisition in the kernel proper anyways.

PR #474, opened 2019-09-16
Secure boot to kernel

roblabla:
Automatically creates a secure boot image that starts at the flash firmware (OVMF) and verifies the bootloader, which in turn verifies the bootstrap, kernel and KIPs.

Long term, we'll replace GRUB (and PGP) with our own bootloader (See #468) but in the meantime, this gets us some very good security memes.


bootstrap/src/main.rs
```diff
@@ -149,10 +149,11 @@ pub extern "C" fn do_bootstrap(multiboot_info_addr: usize) -> ! {
 
     // Move the multiboot_header to a single page in kernel space. This simplifies some
```
Orycterope: Well then it's not a single page, innit ?

Issue #473, opened 2019-09-15
Occasional userland null ptr dereference due to ELF TLS


roblabla:
Sometimes, it looks like gs = 0 when returning to userland, causing null pointer derefs. This seems to happen on current master if you let the VM run for a couple minutes.

Orycterope:
userspace or kernel space ?

either

- `gs` is wrong (i.e. incorrectly swapped. Wouldn't be the first time). Refer `gs >> 3` to https://sunriseos.github.io/SunriseOS/master/sunrise_kernel/i386/gdt/index.html 
- `[gs]` is wrong: Maybe called before elf tls initialization, or I messed up the TLS init image (again, wouldn't be the first time).
Compare `nm <service> | grep __tls` to `readelf --segments`.

roblabla:
@Orycterope userspace is sometimes run with `gs = 0`. We pagefaults on instructions like `mov ecx, gs:[0]` with an invalid read on address 0.

Orycterope:
Wait, `gs` is 0, or `gs` points to 0 ? 

I see no reason for it to be zero

roblabla:
gs points to a segment whose base is 0.

roblabla:
OK scratch that. `gs` is 0. What the fuck.

roblabla:
So I did a quickfix where I changed trap_gate_asm to backup `gs` into UserSpaceHardwareContext and then restore it, and it seems to make the problem go away. I'm a bit ambivalent with the patch, and what bothers me the most is that I don't understand *how* we can end up in this situation. Like, at best I could understand `gs` pointing on the kernel segment, but this makes no sense 🤔 .

roblabla:
With #475 merged, the bug shouldn't cause crashes anymore at least. But it might still cause weirdness in some places (for instance, assuming it is indeed the detection that's broken, esp_cpy might be wrong). Keeping this open until we have a proper investigation of what was the cause of this bug.

PR #475, opened 2019-09-16
Kernel: Fix random userspace gs = 0

roblabla:
Sometimes, when running userspace, GS would be 0, leading to instructions accessing gs:[0] to fail with a null deref. I'm... still not entirely sure why that happens. It appears that our ring detection in trap_gate_asm fails, leading to gs not being changed before returning to userspace. But what confuses me is that when this happens, gs should be set to the kernel TLS, and not 0.

Anyways, this is fixed by pushing GS on the UserSpaceHardwareContext and restoring it unconditionally before returning.

Fixes (but we don't know why) #473

### DIFF
--- a/kernel/src/i386/interrupt_service_routines.rs
+++ b/kernel/src/i386/interrupt_service_routines.rs
@@ -78,6 +78,7 @@ pub fn check_thread_killed() {
 #[allow(missing_docs)]
 pub struct UserspaceHardwareContext {
     pub esp: usize,
+    pub gs: usize,
     pub ebp: usize,
     pub edi: usize,
     pub esi: usize,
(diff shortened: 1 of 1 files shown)


Orycterope:
As you said, the bug makes no sense, and to me finding a way around it without trying to actually understand it, isn't the way to go about it. 

I'd really like to understand the root of the problem, as it might hint toward something being really wrong somewhere. I'll do that asap.

In the mean time, backing up `gs` doesn't really makes sense, it should be UTLS for user, KTLS for user, and never anything else, there's no need for context here.

roblabla:
I agree, but this still fixes the underlying crash which is crippling (at least to me), and makes the logic a bit easier to follow. We might want to open a new issue afterwards to investigate what the real cause is.

Orycterope:
Well ok then, i approve, but keep an issue open

PR #476, opened 2019-09-16
Put min approvals down to 1.

roblabla:


roblabla:
> Could be nice if having one refusal still blocked

Agreed, but mergeable doesn't have an option for it right now, and I don't want to dedicate time to it. Let's operate on a gentleman's agreement :)

PR #484, opened 2019-09-18
Add stack_size as Thread::create argument

Thog:
This makes the stack size used by a thread configurable at creation time. (This is needed for libstd)

### DIFF
--- a/libuser/src/threads.rs
+++ b/libuser/src/threads.rs
@@ -82,13 +82,24 @@ struct StackContext {
 }
 
 impl StackContext {
-    /// Create a new StackContext from a given size.
+    /// Create a new StackContext from a given size. The stack size must be bigger than 0.
+    ///
+    /// # Errors
+    ///
+    /// - `InvalidSize`
+    ///   - The size passed was 0
+    ///   - The size overflows when rounded up to the nearest multiple of PAGE_SIZE.
     pub fn new(stack_size: usize) -> Result<Self, Error> {
-        let stack_layout_res: Result<Layout, Error> = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE).map_err(|_| KernelError::InvalidSize.into());
-        let stack_layout = stack_layout_res?;
+        if stack_size == 0 {
+            return Err(KernelError::InvalidSize.into());
+        }
+
+        let stack_layout = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE)
+            .or(Err(KernelError::InvalidSize))?;
 
         Ok(StackContext {
             stack_address: unsafe {
+                // Safety: We manually alloc the stack from the given stack_layout and dealloc it on Drop.
                 alloc(stack_layout) as *const u8
             },
             stack_layout
@@ -104,11 +115,13 @@ impl StackContext {
 impl Drop for StackContext {
     fn drop(&mut self) {
         unsafe {
+            // Safety: We manually desalloc the stack only when droping the StackContext so this is safe.
             dealloc(self.stack_address as *mut u8, self.stack_layout);
         }
     }
 }
 
+// Safety: This is safe as StackContext content will not be modified after creation.
 unsafe impl Sync for StackContext {}
 unsafe impl Send for StackContext {}
 
--- a/libuser/src/threads.rs
+++ b/libuser/src/threads.rs
@@ -95,6 +95,7 @@ impl StackContext {
         })
     }
 
+    /// Get the address of the stack top.
     pub fn get_stack_top(&self) -> *const u8 {
         self.stack_address.wrapping_add(self.stack_layout.size())
     }
--- a/libuser/src/threads.rs
+++ b/libuser/src/threads.rs
@@ -99,7 +99,7 @@ impl StackContext {
 
         Ok(StackContext {
             stack_address: unsafe {
-                // Safety: We manually alloc the stack from the given stack_layout and dealloc it on Drop.
+                // Safety: We error from the function early if stack_size is 0. We don't care much about whether the block is initialized.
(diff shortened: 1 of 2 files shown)


libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+
+impl Drop for StackContext {
+    fn drop(&mut self) {
+        unsafe {
+            dealloc(self.stack_address as *mut u8, self.stack_layout);
+        }
+    }
+}
+
+unsafe impl Sync for StackContext {}
```
roblabla: Safety comments please.

libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+    stack_address: *const u8,
+
+    /// The stack layout.
+    stack_layout: Layout
+}
+
+impl StackContext {
+    /// Create a new StackContext from a given size.
+    pub fn new(stack_size: usize) -> Result<Self, Error> {
+        let stack_layout_res: Result<Layout, Error> = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE).map_err(|_| KernelError::InvalidSize.into());
```
roblabla:
```suggestion
        let stack_layout = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE)
            .or(KernelError::InvalidSize)?;
```

libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+        let stack_layout = stack_layout_res?;
```
roblabla:
```suggestion
```

libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+struct StackContext {
+    /// The addresss of the allocated stack
+    stack_address: *const u8,
+
+    /// The stack layout.
+    stack_layout: Layout
+}
+
+impl StackContext {
+    /// Create a new StackContext from a given size.
```
roblabla:
```suggestion
    /// Create a new StackContext from a given size. The stack size must be bigger than 0.
    ///
    /// # Errors
    ///
    /// - `InvalidSize`
    ///   - The size passed was 0
    ///   - The size overflows when rounded up to the nearest multiple of PAGE_SIZE.
```

libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+    pub fn new(stack_size: usize) -> Result<Self, Error> {
+        let stack_layout_res: Result<Layout, Error> = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE).map_err(|_| KernelError::InvalidSize.into());
+        let stack_layout = stack_layout_res?;
+
+        Ok(StackContext {
+            stack_address: unsafe {
+                alloc(stack_layout) as *const u8
```
roblabla: You need to check that stack_layout.size() is != 0, otherwise UB will occur! The Layout::new() function does not check for this.

Also, safety comments please.

libuser/src/threads.rs
```diff
@@ -59,15 +59,58 @@
...
+    /// Get the address of the stack top.
+    pub fn get_stack_top(&self) -> *const u8 {
+        self.stack_address.wrapping_add(self.stack_layout.size())
+    }
+}
+
+impl Drop for StackContext {
+    fn drop(&mut self) {
+        unsafe {
+            dealloc(self.stack_address as *mut u8, self.stack_layout);
```
roblabla: Safety comment.

libuser/src/threads.rs
Thog: oops

libuser/src/threads.rs
```diff
@@ -59,15 +59,71 @@
...
+        if stack_size == 0 {
+            return Err(KernelError::InvalidSize.into());
+        }
+
+        let stack_layout = Layout::from_size_align(stack_size, crate::mem::PAGE_SIZE)
+            .or(Err(KernelError::InvalidSize))?;
+
+        Ok(StackContext {
+            stack_address: unsafe {
+                // Safety: We manually alloc the stack from the given stack_layout and dealloc it on Drop.
```
roblabla: This isn't a good safety comment. You need to explain why this piece of unsafe code is sound. The alloc function [documents the situations in which it is unsound](https://doc.rust-lang.org/alloc/alloc/trait.GlobalAlloc.html#safety-1). We need to show that we never hit those cases.

```suggestion
                // Safety: We error from the function early if stack_size is 0. We don't care much about whether the block is initialized.
```

libuser/src/threads.rs
```diff
@@ -59,15 +59,71 @@
...
+    /// Get the address of the stack top.
+    pub fn get_stack_top(&self) -> *const u8 {
+        self.stack_address.wrapping_add(self.stack_layout.size())
+    }
+}
+
+impl Drop for StackContext {
+    fn drop(&mut self) {
+        unsafe {
+            // Safety: We manually desalloc the stack only when droping the StackContext so this is safe.
```
roblabla: Ditto.

```suggestion
            // Safety: The stack_address is guaranteed to be valid (it was allocated on construction). We also keep the layout around to ensure it stays the same between alloc and dealloc.
```

libuser/src/threads.rs
```diff
@@ -59,15 +59,71 @@
...
+            dealloc(self.stack_address as *mut u8, self.stack_layout);
+        }
+    }
+}
+
+// Safety: This is safe as StackContext content will not be modified after creation.
```
roblabla:
```suggestion
// Safety: This is safe as StackContext does not contain any internal mutability.
// In fact, its content (that is, the pointer itself and the layout) are immutable after creation.
```

PR #486, opened 2019-09-20
Implement Thread::join

Thog:
This allows to wait for a thread to be in an exit state.

Also rearrange ThreadState to be closer to Horizon states.

### DIFF
--- a/kernel/src/process.rs
+++ b/kernel/src/process.rs
@@ -211,7 +211,7 @@ pub enum Handle {
     SharedMemory(Arc<SpinRwLock<Vec<PhysicalMemRegion>>>),
 }
 
-/// The underlying shared object of a [ThreadStateReadableEvent].
+/// The underlying shared object of a [Weak<ThreadStrct>].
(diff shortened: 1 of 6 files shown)


roblabla:
Why is state_event arc'd? It doesn't look necessary.

Thog:
updated

kernel/src/process.rs
```diff
@@ -206,6 +211,42 @@ pub enum Handle {
...
+            scheduler::add_to_schedule_queue(thread);
+        }
+    }
+}
+
+/// If this waitable is signaled, this means the thread has exited.
+impl Waitable for Weak<ThreadStruct> {
+    fn is_signaled(&self) -> bool {
+        if let Some(thread) = self.upgrade() {
+            return thread.state.load(Ordering::Relaxed) == ThreadState::TerminationPending;
```
roblabla: I don't think Relaxed works here. This should almost certainly be *at least* Acquire. To avoid trouble, let's just use SeqCst.

PR #489, opened 2019-09-22
Kernel: Use SeqCst for ARE_CPU_LOCALS_INIT

roblabla:
Relaxed is 100% broken in SMP context here.


roblabla:
From @orycterope on discord

> Since we set the atomic to true before we start the other cores, shouldn't Relaxed work?

I'm not 100% sure. I feel like it should work in practice, but I'm very unclear on whether this is considered legal or not.

PR #495, opened 2019-09-25
Fix arguments loading/parsing

Thog:
Thanks I hate it (tm)


roblabla:
TODO should get addressed, and ideally the function should move to libutils

PR #505, opened 2019-10-08
Update libuser Cargo.toml

Thog:
Needed for future std integration


libuser/Cargo.toml
```diff
@@ -33,3 +33,10 @@ version = "0.4.10"
 [dependencies.hashbrown]
 features = ["nightly"]
 version = "0.5.0"
+
+[features]
+default = []
+disable-lang-items = []
```
roblabla: No negative features.

PR #512, opened 2019-10-14
Move terminal handling to Vi

roblabla:
This PR's main goal is to move terminal handling inside VI, but it contains a lot of unrelated improvements. It is best reviewed commit by commit.


roblabla:
This PR makes https://github.com/sunriseos/SunriseOS/issues/81 **much** worse. We now have SharedMemory that mutates **within the same process**. This is clearly UB.

I have an idea floating in my mind to fix it:

I think the way to do this safely is to have `Buffer` be an `[AtomicU8; size]`. Each write should be Relaxed (which basically makes no different from a normal assignment), and once everything is written, add a `fence(SeqCst)`

On the read side, I should `fence(SeqCst)` before doing all reads with Relaxed.

In theory, this solution fixes both cross-process and the new very-UB in-process troubles. Technically speaking, cross-process behavior of atomic isn't spec'd at all. But from what I've read in various places, it's pretty safe to assume that the "core" operations will work in cross-process environment. i386 doesn't really care about the consistency model anyways (it enforces a strong memory model), while on ARM the ops this lowers to will work even in the face of different address spaces.

roblabla:
All fixed and ready to go in theory.

PR #524, opened 2019-10-23
Libstd stdout goes to terminal

roblabla:
We now implement a twili service that's responsible for giving a process its stdin/stdout/stderr. Stderr is ignored for now, as it's used for panics, and having panics go to svcOutputDebugString is too useful.

### DIFF
--- a/libuser/src/syscalls.rs
+++ b/libuser/src/syscalls.rs
@@ -724,7 +724,7 @@ pub(crate) fn reset_signal(event: HandleRef) -> Result<(), KernelError> {
 ///   - The given handle is invalid or not a process.
 pub fn get_process_id(process_handle: &Process) -> Result<u64, KernelError> {
     unsafe {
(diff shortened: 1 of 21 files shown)


libuser/src/syscalls.rs
```diff
@@ -724,7 +724,7 @@ pub(crate) fn reset_signal(event: HandleRef) -> Result<(), KernelError> {
 ///   - The given handle is invalid or not a process.
 pub fn get_process_id(process_handle: &Process) -> Result<u64, KernelError> {
     unsafe {
-        let (pid, ..) = syscall(nr::GetProcessInfo, (process_handle.0).0.get() as usize, 0, 0, 0, 0, 0)?;
+        let (pid, ..) = syscall(nr::GetProcessId, (process_handle.0).0.get() as usize, 0, 0, 0, 0, 0)?;
```
Thog: oops

PR #527, opened 2019-10-26
Keyboard refactor

kitlith:
My goal here was to try to read from the port in as few places as possible (as well as trying to handle weird edge cases) in preparation for sending commands / receiving responses. In the process, though, I may have made `KeyEvent::read_key_event` an even more messy function than before...

### DIFF
--- a/keyboard/src/ps2.rs
+++ b/keyboard/src/ps2.rs
@@ -16,6 +16,7 @@
 
 use sunrise_libuser::io::{Io, Pio};
 use core::sync::atomic::{AtomicBool, Ordering::SeqCst};
+use spin::RwLock;
 use sunrise_libuser::syscalls;
 use sunrise_libuser::types::ReadableEvent;
 
@@ -23,6 +24,7 @@ use sunrise_libuser::keyboard::HidKeyboardState;
 use sunrise_libuser::keyboard::HidKeyboardStateType;
 use sunrise_libuser::keyboard::HidKeyboardScancode;
 use lazy_static::lazy_static;
+use arrayvec::ArrayVec;
 use log::{debug, warn};
 
 /// PS2 keyboard state.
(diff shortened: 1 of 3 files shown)


keyboard/src/ps2.rs
```diff
@@ -113,40 +117,50 @@ struct KeyEvent {
...
+        //let scancode = port.read();
         let mut state = Pressed;
 
-        let key = match scancode {
+        let key = match input.get(0).expect("How did this get called with 0 scancodes?") {
 
             // multibyte scancodes
             0xe0 => {
-                match port.read() {
+                match input.get(1) {
```
roblabla: To make this code a bit cleaner, it'd be a good idea to `impl From<NoneError> for KeyEventError`, and return KeyEventError. Then we could just have `input.get(1)?` to handle the error case automatically. Though I suppose that was the intent and you just didn't get to it yet.

kitlith: Mm, that wouldn't be a bad idea. I emded up not using the error type because I realized that unknowns were already handled, so there's only one type of error, 'incomplete'... we could also change it to an option since ? works with options. (even got a lint about that from clippy on my first push)

kitlith:
It's not exactly something that I directly had in mind, but I'm glad it'll be helpful for that too.

Orycterope:
Regarding moving key reading to a separate process, I'm not sure what you meant, but I guess a producer-consumer pattern would fit. 

One task (producer) waits for interrupts, fills a ring buffer with every byte it can read, and then goes to sleep again. It does not attempt to interpret read bytes in anyway, just bufferize them. 
Another task (consumer) waits for IPC calls, and tries to interpret the content of the buffer as scancodes, and goes to sleep if it's empty. 

This would require locks to synchronize the two tasks, which we don't have yet (or I missed them). 

Anyway this is irrelevant to the current PR, which is fine ! Thank you, your contribution is really appreciated ! 😊

roblabla:
Needs a rebase on master to fix the build failure.

kitlith:
I wonder if the pattern matching could be simplified with slice patterns nowadays...

PR #528, opened 2019-10-27
 Panic if SpinLock locked in interrupt context

### LINKED ISSUE #471: Detect usage of SpinLock in IRQ context
It'd be nice if we could detect use of a SpinLock in an IRQ context. Such usage would be wrong (since it might end up deadlocking the CPU). @Orycterope I wonder if we could set a global boolean or something in an IRQ, and check it when locking a SpinLock, panicking if it's set?

kitlith:
Fixes #471.

This involves creating a newtype wrapper around spin::Mutex to check when locking if the number of interrupts being serviced is currently zero.

### DIFF
--- a/kernel/src/sync/spin_lock.rs
+++ b/kernel/src/sync/spin_lock.rs
@@ -121,7 +121,10 @@ impl<T: ?Sized> SpinLock<T> {
     /// ```
     pub fn lock(&self) -> SpinLockGuard<T> {
         use core::sync::atomic::Ordering;
-        if crate::i386::interrupt_service_routines::INSIDE_INTERRUPT_COUNT.load(Ordering::SeqCst) != 0 {
+        use crate::cpu_locals::ARE_CPU_LOCALS_INITIALIZED_YET;
+        use crate::i386::interrupt_service_routines::INSIDE_INTERRUPT_COUNT;
+        use super::INTERRUPT_DISARM;
+        if !INTERRUPT_DISARM.load(Ordering::SeqCst) && ARE_CPU_LOCALS_INITIALIZED_YET.load(Ordering::SeqCst) && INSIDE_INTERRUPT_COUNT.load(Ordering::SeqCst) != 0 {
             panic!("\
                 You have attempted to lock a spinlock in interrupt context. \
                 This is most likely a design flaw. \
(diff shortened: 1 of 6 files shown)


kitlith:
It looks like maybe I should fiddle with this a bit more *if* syscall context is not considered irq context. (Is it? It's the only 'interrupt' where interrupts are not disabled.) maybe add an extra thing to the macro?

Orycterope:
Indeed, syscalls context isn't interrupt context 😕

kernel/src/sync/spin_lock.rs
```diff
@@ -0,0 +1,156 @@
...
+    ///     let mut data = mylock.lock();
+    ///     // The lock is now locked and the data can be accessed
+    ///     *data += 1;
+    ///     // The lock is implicitly dropped
+    /// }
+    ///
+    /// ```
+    pub fn lock(&self) -> SpinLockGuard<T> {
+        use core::sync::atomic::Ordering;
+        assert_eq!(crate::i386::interrupt_service_routines::INSIDE_INTERRUPT_COUNT.load(Ordering::SeqCst), 0);
```
Orycterope: Add a verbose panic message telling the user it has a design flaw and should refer to the sync module documentation.

kernel/src/sync/spin_lock.rs
```diff
@@ -0,0 +1,156 @@
...
+    ///
+    /// If the lock isn't held, this is a no-op.
+    pub unsafe fn force_unlock(&self) {
+        self.0.force_unlock()
+    }
+
+    /// Tries to lock the spinlock. If it is already locked, it will return None. Otherwise it returns
+    /// a guard within Some.
+    pub fn try_lock(&self) -> Option<SpinLockGuard<T>> {
+        self.0.try_lock()
```
Orycterope: `try_lock` should panic too. The user made a design flaw, be loud about it.

kitlith:
Thanks for the feedback, I'll try to get to that today. Fixing syscall context in particular is going to be annoying (but that'll also lay the groundwork for switching back to counter-based SpinLockIRQ, i think.)

roblabla:
LGTM, can get merged after a rebase :)

PR #534, opened 2019-10-31
Switch back to Counter based SpinLockIRQ

### LINKED ISSUE #190: SpinLockIRQ is totally broken
SpinLockIRQ doesn't fulfill its intended purpose. In the interrupt handler, when we drop the handler, we end up reactivating interruptions before returning. This leads to a kernel stackoverflow. We should revert to the previous "save and restore" design. However, something needs to be done to prevent ordering problems.

# The reordering problem

The traditional way to handle SpinLockIRQ is to have lock() backup the EFLAGS (in particular, the IF flag, which prevents interruptions), and then disable interrupts. Unlock restores IF. This works fine, so long as the locks are unlocked in reverse order as they were locked. Ordering them wrong will lead to terrible consequences. In particular, the first lock should be the last unlock. Failure to do this will result in the EFLAGS being left in an invalid state.

# Potential solutions

In our case, lock() returns a Guard, and unlock() occurs when the Guard is dropped. The reordering problem occurs if we move the guard to a subfunction, or manually drop it. In both cases, this would require a move.

Rust has a new API, called Pin, that is made to prevent moves: https://doc.rust-lang.org/std/pin/struct.Pin.html . Perhaps we can use this to prevent the guards from being moved around, and as such ensure the correct drop order is respected?

### LINKED ISSUE #192: Find sane design for SpinLockIRQ safety
Currently, SpinLockIRQ API is unsound. If the guards are dropped in the wrong order, it may cause IF to be reset too early.
Ideally, we would need a way to prevent the guard variable to be reassigned. AKA: prevent moving. Note that this is different from what the Pin API solves. The Pin API is about locking a variable in one memory location, but its binding may still be moved and dropped. Unfortunately, Rust does not have a way to express that a value cannot be reassigned.
Another possibility would be to switch to a callback API. This would solve the problem, but the scheduler would be unable to consume such locks. Maybe we could have an unsafe &quot;scheduler_relock&quot; function that may only be called from the scheduler?

---

https://github.com/roblabla42/KFS/blob/e850a8908338a52d5f8e185a5bd914f669b615e1/kernel/src/sync.rs#L53-L63

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in e850a8908338a52d5f8e185a5bd914f669b615e1 when #191 was merged. cc @roblabla.

kitlith:
Fixes #190. Fixes #192.

# Why this doesn't suck
Interrupt handler wrappers are now generated by a macro, so we have a single place to inject code into interrupt handlers. This allows us to tell SpinLockIRQ that interrupts are already disabled by incrementing the counter and allows us to decrement the counter again (without disabling interrupts yet) right before entering userspace again.

This also contains a fix for a mistake in #528, because I failed to copy-paste some checks into try_lock. >_> If this PR doesn't go through, then that should still be fixed anyway.

### DIFF
--- a/kernel/src/i386/interrupt_service_routines.rs
+++ b/kernel/src/i386/interrupt_service_routines.rs
@@ -1137,5 +1137,5 @@ pub unsafe fn init() {
         (*idt).load();
     }
 
-    sti();
+    crate::sync::spin_lock_irq::enable_interrupts();
 }
--- a/kernel/src/i386/interrupt_service_routines.rs
+++ b/kernel/src/i386/interrupt_service_routines.rs
@@ -567,10 +567,12 @@ macro_rules! generate_trap_gate_handler {
 
             use crate::i386::structures::gdt::SegmentSelector;
             use crate::i386::interrupt_service_routines::INSIDE_INTERRUPT_COUNT;
+            use crate::sync::spin_lock_irq::{disable_interrupts, decrement_lock_count};
             use core::sync::atomic::Ordering;
 
             if $interrupt_context {
                 let _ = INSIDE_INTERRUPT_COUNT.fetch_add(1, Ordering::SeqCst);
+                disable_interrupts();
             }
 
             if let PrivilegeLevel::Ring0 = SegmentSelector(userspace_context.cs as u16).rpl() {
@@ -592,6 +594,7 @@ macro_rules! generate_trap_gate_handler {
 
             if $interrupt_context {
                 let _ = INSIDE_INTERRUPT_COUNT.fetch_sub(1, Ordering::SeqCst);
+                unsafe { decrement_lock_count(); }
             }
 
             // if we're returning to userspace, check we haven't been killed
--- a/kernel/src/i386/process_switch.rs
+++ b/kernel/src/i386/process_switch.rs
@@ -347,6 +347,8 @@ fn jump_to_entrypoint(ep: usize, userspace_stack_ptr: usize, arg1: usize, arg2: 
     const_assert_eq!((GdtIndex::UTlsRegion as u16) << 3 | 0b11, 0x3B);
     const_assert_eq!((GdtIndex::UTlsElf as u16) << 3 | 0b11, 0x43);
     const_assert_eq!((GdtIndex::UStack as u16) << 3 | 0b11, 0x4B);
+    // We're about to enable interrupts by doing an iret.
+    unsafe { crate::sync::spin_lock_irq::decrement_lock_count(); }
     unsafe {
         asm!("
         mov ax,0x33  // ds, es <- UData, Ring 3
--- a/kernel/src/sync/spin_lock_irq.rs
+++ b/kernel/src/sync/spin_lock_irq.rs
@@ -11,6 +11,7 @@ use core::mem::ManuallyDrop;
 use core::ops::{Deref, DerefMut};
 use core::sync::atomic::{AtomicU8, Ordering};
 use super::INTERRUPT_DISARM;
+use crate::cpu_locals::ARE_CPU_LOCALS_INITIALIZED_YET;
 
 /// Interrupt disable counter.
 ///
@@ -28,7 +29,7 @@ static INTERRUPT_DISABLE_COUNTER: AtomicU8 = AtomicU8::new(0);
 ///
 /// Look at documentation for ProcessStruct::pint_disable_counter to know more.
 pub fn enable_interrupts() {
-    if !INTERRUPT_DISARM.load(Ordering::SeqCst) {
+    if !INTERRUPT_DISARM.load(Ordering::SeqCst) && ARE_CPU_LOCALS_INITIALIZED_YET.load(Ordering::SeqCst) {
         if INTERRUPT_DISABLE_COUNTER.fetch_sub(1, Ordering::SeqCst) == 1 {
             unsafe { interrupts::sti() }
         }
@@ -40,7 +41,7 @@ pub fn enable_interrupts() {
 /// Used to decrement counter while keeping interrupts disabled inside an interrupt.
 /// Look at documentation for INTERRUPT_DISABLE_COUNTER to know more.
 pub unsafe fn decrement_lock_count() {
-    if !INTERRUPT_DISARM.load(Ordering::SeqCst) {
+    if !INTERRUPT_DISARM.load(Ordering::SeqCst) && ARE_CPU_LOCALS_INITIALIZED_YET.load(Ordering::SeqCst) {
         let _ = INTERRUPT_DISABLE_COUNTER.fetch_sub(1, Ordering::SeqCst);
     }
 }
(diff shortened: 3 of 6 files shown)


kernel/src/i386/interrupt_service_routines.rs
```diff
@@ -592,6 +594,7 @@ macro_rules! generate_trap_gate_handler {
 
             if $interrupt_context {
                 let _ = INSIDE_INTERRUPT_COUNT.fetch_sub(1, Ordering::SeqCst);
+                unsafe { decrement_lock_count(); }
```
roblabla: unsafe needs to have a `// Safety` comment, [like this](https://github.com/sunriseos/SunriseOS/blob/d05f23d5a4816d8ca493d587d12ba1b379f0e523/libuser/src/argv.rs#L121)

kernel/src/i386/process_switch.rs
```diff
@@ -347,6 +347,8 @@ fn jump_to_entrypoint(ep: usize, userspace_stack_ptr: usize, arg1: usize, arg2:
     const_assert_eq!((GdtIndex::UTlsRegion as u16) << 3 | 0b11, 0x3B);
     const_assert_eq!((GdtIndex::UTlsElf as u16) << 3 | 0b11, 0x43);
     const_assert_eq!((GdtIndex::UStack as u16) << 3 | 0b11, 0x4B);
+    // We're about to enable interrupts by doing an iret.
+    unsafe { crate::sync::spin_lock_irq::decrement_lock_count(); }
```
roblabla: IDEM, needs a `// Safety` comment explaining why this operation is safe.

Also, I'm a bit confused at why this is necessary? Is this what was talked about on discord, because of the `interrupt_manager` SpinLockIRQ? If so:

1. It should definitely have a big fat comment explaining this in detail
2. I think we can get rid of `interrupt_manager`

kernel/src/sync/spin_lock_irq.rs
```diff
@@ -74,45 +105,32 @@ impl<T> SpinLockIRQ<T> {
...
+
+        // TODO: Disable preemption.
+        // TODO: Spin acquire
+
+        // lock
+        match self.internal.try_lock() {
+            Some(internalguard) => Some(SpinLockIRQGuard(ManuallyDrop::new(internalguard))),
+            None => {
+                // We couldn't lock. Restore irqs and return None
+                unsafe { enable_interrupts(); }
```
roblabla: Safety comment

kernel/src/sync/spin_lock_irq.rs
```diff
@@ -74,45 +105,32 @@ impl<T> SpinLockIRQ<T> {
...
-                // if lock is successful, return guard.
-                Some(SpinLockIRQGuard(ManuallyDrop::new(internalguard), saved_intpt_flag))
-            } else {
-                // Else, restore interrupt state
-                if saved_intpt_flag {
-                    unsafe { interrupts::sti(); }
-                }
+    pub fn try_lock(&self) -> Option<SpinLockIRQGuard<'_, T>> {
+        // Disable irqs
+        unsafe { disable_interrupts(); }
```
roblabla: Safety comment

kernel/src/sync/spin_lock_irq.rs
```diff
@@ -74,45 +105,32 @@ impl<T> SpinLockIRQ<T> {
...
-
-            // Disable interruptions
-            unsafe { interrupts::cli(); }
-
-            let internalguard = self.internal.lock();
-            SpinLockIRQGuard(ManuallyDrop::new(internalguard), saved_intpt_flag)
-        }
+    pub fn lock(&self) -> SpinLockIRQGuard<'_, T> {
+        // Disable irqs
+        unsafe { disable_interrupts(); }
```
roblabla: Safety comment

kernel/src/i386/process_switch.rs
kitlith: why is interrupt manager necessary?

the guard for the queue is dropped before the context switch, which must still have interrupts disabled. So it's not removable as such.

Alternatives to what I *did* do include wrapping context_switch in decrements/increments of the counter and/or removing the guard object and using calls to increment/decrement. I think it's a bit better to associate it with the iret, but *shrug*.

kitlith:
They don't, per-se. Even the one which I have marked `unsafe` consists of entirely safe code. (I haven't marked the other two as unsafe because they were previously safe, but I am willing to change that.)

However, I consider calling them unsafe, as manually incrementing/decrementing the counter at the wrong time is going to break the counter and cause deadlocks.

Could possibly make it less fragile by disabling interrupts every time disable_interrupts is called, but that probably means that stuff will just break in a different, non-obvious way.

kernel/src/i386/process_switch.rs
kitlith: Oh, here's another alternative: making context-switch and co less unsafe by having it manage interrupt enabling/disabling by itself instead of making everything else manage it.

roblabla:
The safety comments on the `unsafe fn` should be modeled [like clippy asks](https://rust-lang.github.io/rust-clippy/master/#missing_safety_doc) (clippy version used in SunriseOS doesn't have that lint right now, but we'll update soon-ish). E.G. it should be a `# Safety` section in the markdown (and not `# Unsafety`).

As for the `// Safety` comments, they should go *inside* the unsafe block. I'm planning on making a tool that looks for unsafe blocks and checks that they contain a safety comment ^^'.

kernel/src/i386/interrupt_service_routines.rs
```diff
@@ -346,6 +349,10 @@ macro_rules! trap_gate_asm {
 ///     * `ignore`: don't do anything for this interrupt.
 ///     * `kill`: kills the process in which this interrupt originated.
 ///     * `my_handler_func`: calls `my_handler_func` to handle this interrupt. Useful if you want to override a standard strategy.
+/// * The possible values for `type` are:
+///     * `syscall`: interrupts and usage of regular SpinLocks are allowed.
+///     * `exception`: interrupts are disabled, but regular SpinLocks can be accessed if coming from userspace.
```
roblabla:
```suggestion
///     * `exception`: interrupts are disabled, but regular SpinLocks can be accessed if the exception comes userspace.
```

kernel/src/i386/interrupt_service_routines.rs
```diff
@@ -346,6 +349,10 @@ macro_rules! trap_gate_asm {
 ///     * `ignore`: don't do anything for this interrupt.
 ///     * `kill`: kills the process in which this interrupt originated.
 ///     * `my_handler_func`: calls `my_handler_func` to handle this interrupt. Useful if you want to override a standard strategy.
+/// * The possible values for `type` are:
+///     * `syscall`: interrupts and usage of regular SpinLocks are allowed.
```
roblabla:
```suggestion
///     * `syscall`: interrupts are enabled, and usage of regular SpinLocks are allowed.
```

PR #537, opened 2019-11-01
Implement df

Thog:
Because KFS-X needs it


df/src/main.rs
```diff
@@ -0,0 +1,88 @@
+use std::os::sunrise::prelude::*;
```
roblabla:
```suggestion
// rustc warnings
#![warn(unused)]
#![warn(missing_debug_implementations)]
#![allow(unused_unsafe)]
#![allow(unreachable_code)]
#![allow(dead_code)]
#![cfg_attr(test, allow(unused_imports))]

// rustdoc warnings
#![warn(missing_docs)] // hopefully this will soon become deny(missing_docs)
#![deny(intra_doc_link_resolution_failure)]

use std::os::sunrise::prelude::*;
```

Thog: oops

PR #544, opened 2019-11-22
Implement environment variables

roblabla:
Simple.


std_hello_world/src/main.rs
```diff
@@ -88,6 +94,7 @@ fn main() {
         println!("Child process is dead!");
     }
 
+
```
Thog: Extra space not needed

PR #562, opened 2020-03-10
CI: Bitrot protection for dep installs.

roblabla:
Ensure we build xargo and cargo-make in a reproducible way to prevent bitrot. We now specify a version and ask for a locked build. My understanding is that --locked will use the lockfile packaged by the crate, see https://github.com/rust-lang/cargo/issues/5654.


roblabla:
Sneaky merge

PR #566, opened 2020-03-20
Add partial aarch64 support to libuser

leo60228:
This is able to make libuser compile with the included target JSON. It does not yet include a crt0 or syscall stub, though preparations for that have been made (I'm not experienced enough with aarch64 assembly).

### DIFF
--- a/libuser/src/syscalls.rs
+++ b/libuser/src/syscalls.rs
@@ -83,25 +83,36 @@ extern {
     fn syscall_inner(registers: &mut Registers);
 }
 
+// unsafe fn syscall(nr: usize, arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize, arg6: usize) -> Result<(usize, usize, usize, usize), KernelError> {
 /// Generic syscall function.
-unsafe fn syscall(nr: usize, arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize, arg6: usize) -> Result<(usize, usize, usize, usize), KernelError> {
-    let mut registers = Registers {
-        eax: nr,
-        ebx: arg1,
-        ecx: arg2,
-        edx: arg3,
-        esi: arg4,
-        edi: arg5,
-        ebp: arg6
-    };
+macro_rules! syscall {
+    ($nr:expr, $arg1:expr, $arg2:expr, $arg3:expr, $arg4:expr, $arg5:expr, $arg6:expr) => {{
+        let nr: usize = $nr;
+        let arg1: usize = $arg1;
+        let arg2: usize = $arg2;
+        let arg3: usize = $arg3;
+        let arg4: usize = $arg4;
+        let arg5: usize = $arg5;
+        let arg6: usize = $arg6;
 
-    syscall_inner(&mut registers);
+        let mut registers = Registers {
+            eax: nr,
+            ebx: arg1,
+            ecx: arg2,
+            edx: arg3,
+            esi: arg4,
+            edi: arg5,
+            ebp: arg6
+        };
 
-    if registers.eax == 0 {
-        Ok((registers.ebx, registers.ecx, registers.edx, registers.esi))
-    } else {
-        Err(KernelError::from_syscall_ret(registers.eax as u32))
-    }
+        syscall_inner(&mut registers);
+
+        if registers.eax == 0 {
+            Ok((registers.ebx, registers.ecx, registers.edx, registers.esi))
+        } else {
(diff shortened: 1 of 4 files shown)


leo60228:
I've implemented a syscall stub, it was easier than I thought.

libuser/src/syscalls.rs
```diff
@@ -70,41 +70,72 @@ syscall_inner:
...
+        if registers.nr == 0 {
+            Ok((registers.arg1, registers.arg2, registers.arg3, registers.arg4))
+        } else {
+            Err(KernelError::from_syscall_ret(registers.nr as u32))
+        }
+    }};
 }
 
-/// Resize the heap of a process, just like a brk.
+/// Rarg4ze the heap of a process, just like a brk.
```
roblabla: Lol, I too Rarg4ze the heap of a process :P.

libuser/src/syscalls.rs
```diff
@@ -177,29 +222,29 @@ pub fn start_thread(thread_handle: &Thread) -> Result<(), KernelError> {
...
+        syscall!(nr::SleepThread, nanos, 0, 0, 0, 0, 0)?;
         Ok(())
     }
 }
 
 /// Sets the "signaled" state of an event. Calling this on an unsignalled event
 /// will cause any thread waiting on this event through [wait_synchronization()]
 /// to wake up. Any future calls to [wait_synchronization()] with this handle
-/// will immediately return - the user has to clear the "signaled" state through
+/// will immarg5ately return - the user has to clear the "signaled" state through
```
roblabla: here

libuser/src/syscalls.rs
```diff
@@ -515,7 +560,7 @@ pub fn map_mmio_region(physical_address: usize, size: usize, virtual_address: us
 ///
 /// ## x86_64
 ///
-/// ![same, but different, but still same](https://media.giphy.com/media/C6JQPEUsZUyVq/giphy.gif)
+/// ![same, but different, but still same](https://marg5a.giphy.com/marg5a/C6JQPEUsZUyVq/giphy.gif)
```
roblabla: here

libuser/src/syscalls.rs
```diff
@@ -526,11 +571,11 @@ pub fn map_mmio_region(physical_address: usize, size: usize, virtual_address: us
 ///
 /// # Errors
 ///
-/// * The whole initial design of TLS on x86 should be considered an error.
+/// * The whole initial darg4gn of TLS on x86 should be considered an error.
```
roblabla: here

libuser/src/syscalls.rs
```diff
@@ -70,41 +70,72 @@ syscall_inner:
...
+            arg4,
+            arg5,
+            arg6
+        };
+
+        #[cfg(not(target_arch = "aarch64"))]
+        syscall_inner(&mut registers);
+
+        #[cfg(target_arch = "aarch64")]
+        asm!(
```
roblabla: this needs to be a volatile asm block, and needs a memory clobber.

libuser/src/threads.rs
```diff
@@ -193,6 +193,7 @@ static MAIN_THREAD_CONTEXT: ThreadContext = ThreadContext {
 
 /// Get a pointer to this thread's [TLS] region pointed to by `fs`, translated to the flat-memory model.
 #[inline]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
roblabla: why x86_64? This is almost surely wrong. x86_64 TLS works slightly differently iirc.

leo60228: Previously, this was used on all architectures. Docs fail if it isn't defined at all.

libuser/src/syscalls.rs
```diff
@@ -105,8 +105,28 @@ macro_rules! syscall {
             arg6
         };
 
+        #[cfg(not(target_arch = "aarch64"))]
         syscall_inner(&mut registers);
 
+        #[cfg(target_arch = "aarch64")]
+        asm!(
+            "svc $7"
+            : "={x0}"(registers.nr),
```
Orycterope: wow, you did exactly what I was discussing in the previous comment 😅 

For aarch64, I don't see the need for copying arguments to the `Registers` struct. Can't you just directly use `nr` ?

leo60228: I thought Registers was in the public API? If it isn't, that's unnecessary.

Orycterope: It was, but I think that's a mistake, it should be defined only for i386

PR #587, opened 2020-05-30
New rust-lang/rust workflow based on patches

### LINKED ISSUE #583: Better rust-lang/rust integration strategy
Currently, we have a nearly full copy of the rust-lang repo in here. This is undesirable for multiple reason. It makes the repo heavier than it needs to be, and brings a ton of code that is actually unreachable. Besides, we only have very light patches to libstd, and a ton of "new" code for the `libstd::sys`.

I think we should move to a patch-based system. The idea is, we would only store a single (or maybe a couple) `.patch` files in the repo, that would get applied by the build system to the `rust-src` component. `cargo make qemu` would have a new step, "fork-rust", that would `cp -r` the rust-src module into rust/, and apply the patchfiles on top of it. A new command, `cargo make create-patch`, would automatically re-create the patchfile by running `diff rust/ $(rust-src location)`

Several benefits: 

1. It removes the current scenario where we have two sources of trust for the rust fork (sunriseos/rust and sunriseos/sunrise/rust). The fork could be kept to easily see the result of the patch, but be automatically updated through github actions, to prevent it from going out of sync.
2. Updating to latest nightly becomes a fairly obvious and simple business: Just update the patches to make them work. No need to mess around with branches or whatever.

roblabla:
Fixes #583


roblabla:
This needs a few improvements before I'm confident with it. Currently, setup-rust will silently do nothing if the `rust` folder already exists. That's fairly conservative, which is a good property, but there is one case that I'd like to support before going forward: The user has done no changes to their `rust` repo, and the patches got updated. In this case, we should update their rust repo to the latest patches.

The other case to think about: The user has changes to their rust repo, and the patches got updated. In this case, the ideal thing to do would be to print a warning to the user telling them they need to somehow merge their patches with upstream's.

The difficulty here is that it's hard to detect if the patches got updated, especially since the user might cause a patch to change by amending an existing commit. We can probably detect unchanged patches through authorid shenanigans.

sagiegurari:
@roblabla i wonder if instead of 
dependencies = ["install-xargo", "setup-rust"]
everywhere, you might want to do either
* have an init task that invokes those 2
* have 1 task that everyone depends on, and that task depends on those 2. 

both will enable you to make that basic dependencies list grow without having to change all tasks each time (ya, i know its probably rare, but still).

by the way, happy to see that duckscript was able to provide you of a cross platform scripting solution :)
it looks great. thanks for the bug reports.

